PR #3, opened 2023-11-15
RSPY37-38: Prefect CADIP ingestion

Padeanu:


Padeanu:
[RSPY37]: An operator or a developer is able to import a Python function that queries a list of files from a CADIP interface and use it in a Prefect workflow of his own.
[RSPY38]: An operator or a developer is able to import a Python function that queries a list of sessions from a CADIP interface and use it in a Prefect workflow of his own.

Tasks joined since querying files and sessions can be integrated in one continuous flow.
Test to be updated in order to assure compliance with DoD "Check that the error is displayed in Prefect UI but that credentials are not logged." -> Somehow prefect logs and prefect ui should/can be integrated with pytest? to be verified.

PR #9, opened 2023-12-05
Feat rspy32/complete dev ci

jgaucher-cs:


tfache-cs (approved):
it's so awesome !! <3

PR #15, opened 2023-12-20
Feat rspy 35/36 download upload from to s3

agrosu1978:
Merge the refactorized s3 storage handler code to develop,


tests/s3_storage_handler_test.py
jgaucher-cs, line 1: Could we: 

- Rename this file into `test_s3_storage_handler.py` (always put the `_test` at the front because this is what is done in most projects)
- Move the `s3.yml`, `full_s3...` and `short_s3...` file and dirs into a `resources/s3` dir to separate python files and resources

I've done it here: [2fd46321f56c689c301b1b5d791e5e063cb70a38](https://github.com/RS-PYTHON/rs-server/commit/2fd46321f56c689c301b1b5d791e5e063cb70a38)

rs_server/s3_storage_handler/s3_storage_handler.py
```diff
@@ -17,243 +24,719 @@
...
+            Creates a list of local files to be uploaded.
+
+        list_s3_files_obj(self, bucket, prefix):
+            Retrieves the content of an S3 directory.
+
+        get_s3_data(s3_url):
+            Parses S3 URL to extract bucket, prefix, and file.
+
+        check_bucket_access(self, bucket):
+            Checks the accessibility of an S3 bucket.
```
jgaucher-cs, line 74: This "Method" section is useless (because the methods docstrings are available below) and will be too difficult to maintain (i.e. when someone will implement a new method, they will forget to add it in the "Method" section). I suggest we remove it.

rs_server/s3_storage_handler/s3_storage_handler.py
```diff
@@ -17,243 +24,719 @@
...
+            endpoint_url (str): The endpoint URL for the S3 service.
+            region_name (str): The region name.
+
+        Raises:
+            RuntimeError: If the connection to the S3 storage cannot be established.
+        """
+        self.logger = logging.getLogger("s3_storage_handler_test")
+        self.logger.setLevel(logging.DEBUG)
+        self.logger.handlers = []
+        self.logger.addHandler(logging.StreamHandler(sys.stdout))
```
jgaucher-cs, line 92: I think it's better to use a common logging class to avoid code duplication. I've done it here: https://github.com/RS-PYTHON/rs-server/commit/57ae0effdd6f9a617a1cb2a70f21448908bdb0dc

So instead of 
```
    logger = logging.getLogger("s3_storage_handler_test")
    logger.setLevel(logging.DEBUG)
    logger.handlers = []
    logger.addHandler(logging.StreamHandler(sys.stdout))
```
We only do:
```
    logger = Logging.default(__name__)
```

rs_server/s3_storage_handler/s3_storage_handler.py
```diff
@@ -17,243 +24,719 @@
...
+                    endpoint_url=endpoint_url,
+                    region_name=region_name,
+                    config=client_config,
+                )
+            except ClientError as e:
+                if e.response["Error"]["Code"] == "EntityAlreadyExists":
+                    self.logger.error("This clent already exists")
+                else:
+                    self.logger.error("Unexpected error")
+                return None
```
jgaucher-cs, line 139: In modern code, in case of error, we should not return an boolean but raise an Exception. So everyone does not have to check for return values. I've done it here: https://github.com/RS-PYTHON/rs-server/commit/11c017e1fb155a52e83d361febbd9676dd471088

Note that instead of
```
            except ClientError as e:
                if e.response["Error"]["Code"] == "EntityAlreadyExists":
                    self.logger.error("This clent already exists")
                else:
                    self.logger.error("Unexpected error")
```
You can do
```
            except ClientError as e:
                if e.response["Error"]["Code"] == "EntityAlreadyExists":
                    raise RuntimeError("This clent already exists") from e
                else:
                    raise e
```
So you always raise an exception and keep the original exception info.

rs_server/s3_storage_handler/s3_storage_handler.py
```diff
@@ -17,243 +24,719 @@
...
+        """Get the filename from a full path.
+
+        Args:
+            int_path (str): The full path.
+
+        Returns:
+            str: The filename.
+        """
+        path, filename = ntpath.split(input_path)
+        return filename or ntpath.basename(path)
```
jgaucher-cs, line 244: What is the use of this function instead of just using `os.path.basename(input_path)` ?

rs_server/s3_storage_handler/s3_storage_handler.py
```diff
@@ -17,243 +24,719 @@
...
+
+    def disconnect_s3(self):
+        """Close the connection to the S3 service."""
+        if self.s3_client is not None:
+            self.s3_client.close()
+        self.s3_client = None
+
+    # delete a file from s3 by using the s3 handler
+    def delete_file_from_s3(self, bucket, s3_obj):
+        """Delete a file from S3.
```
jgaucher-cs, line 168: You should not have both a python comment and a docstring. You should remove the python comment and put it in the docstring e.g.
```
    def delete_file_from_s3(self, bucket, s3_obj):
        """delete a file from s3 by using the s3 handler.
```
This is done several times in your code.

rs_server/s3_storage_handler/s3_storage_handler.py
```diff
@@ -17,243 +24,719 @@
...
+            with self.s3_client_mutex:
+                self.logger.debug("{%s} | {%s}", bucket, s3_obj)
+                self.logger.info("Delete file s3://{%s}/{%s}", bucket, s3_obj)
+                self.s3_client.delete_object(Bucket=bucket, Key=s3_obj)
+        except ClientError as e:
+            tb = traceback.format_exc()
+            self.logger.error("Failed to delete file s3://{%s}/{%s}", bucket, s3_obj)
+            self.logger.error("Exception: {%s} | {%s}", e, tb)
+            return False
+        return True
```
jgaucher-cs, line 190: Here and in other places, also raise Exceptions instead of returning booleans. I've done it here: https://github.com/RS-PYTHON/rs-server/commit/974fe8ba79c6fc378733982946230feae8509a94

tests/s3_storage_handler_test.py
```diff
@@ -0,0 +1,782 @@
...
+    """
+    assert expected_res == S3StorageHandler.get_basename(path)
+
+
+@pytest.mark.unit
+@pytest.mark.parametrize(
+    "endpoint, bucket, nb_of_files",
+    [
+        (("http://localhost:5000", "test-bucket", 2012)),
+        (("http://localhost:5000", "test-bucket", 999)),
```
jgaucher-cs, line 183: Could we use less files so the test takes less time ? 21 and 9 files would be sufficient instead of 2012 and 999 ?

rs_server/s3_storage_handler/s3_storage_handler.py
agrosu1978: Removed it

rs_server/s3_storage_handler/s3_storage_handler.py
agrosu1978: Merged with the branch you added the exceptions

tests/s3_storage_handler_test.py
agrosu1978: Merged with the branch you did it already

rs_server/s3_storage_handler/s3_storage_handler.py
agrosu1978: Merged with you branch

rs_server/s3_storage_handler/s3_storage_handler.py
agrosu1978: I have had this function for a long time, carrying it with me everywhere. It takes care in case of input_path ends with /. (it's also crossover, win and linux). I would prefer to leave it like this. Just make a test with get_basename("/test/t1/") and os.path.basename("/test/t1/") . The result is different -> t1 with my func, vs empty string with what you are suggesting

rs_server/s3_storage_handler/s3_storage_handler.py
agrosu1978: Solved

rs_server/s3_storage_handler/s3_storage_handler.py
agrosu1978: Merged with your branch

tests/s3_storage_handler_test.py
agrosu1978: No, this is exactly the point of this test. The paginator can handle a max of 1000 files

PR #18, opened 2024-01-15
CADU Download improvement

Padeanu:
This branch was created in order to remove product ID from download endpoint call, small improvements in code and tests and clean up. No major functional changes added.

### DIFF
--- a/rs_server/CADIP/api/cadu_download.py
+++ b/rs_server/CADIP/api/cadu_download.py
@@ -38,7 +38,12 @@ CONF_FOLDER = Path(osp.realpath(osp.dirname(__file__))).parent.parent.parent / "
 logger = Logging.default(__name__)
 
 
-def update_db(db, db_product: CaduDownloadStatus, estatus: EDownloadStatus, status_fail_message=None):
+def update_db(
+    db,
+    db_product: CaduDownloadStatus,
+    estatus: EDownloadStatus,
+    status_fail_message=None,
+):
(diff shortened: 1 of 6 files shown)


rs_server/CADIP/api/cadu_download.py
```diff
@@ -223,10 +226,10 @@ def download(
         db_product = CaduDownloadStatus.get(db, name=name)
     except HTTPException as exception:
```
jgaucher-cs, line 227: Here CaduDownloadStatus.get would raise a HTTPException 404 if the product is missing from the database (don't know if it could happen here). If we lost connection, it would rather raise a sqlalchemy.exc.OperationalError.

(I don't know if we should document all the exceptions that could be raised by all the functions. It would be difficult to maintain this docstring).

But anyway below you're raising a HTTP_503_SERVICE_UNAVAILABLE in case of HTTPException 404, I don't know if it's relevant, and you're not catching sqlalchemy.exc.OperationalError. But do we want to make the effort of catching all exceptions and returning {"started": "false"} ? Can't we let the exception propagate, let the caller catch it and set {"started": "false"} in case of any exception ?

Padeanu: Agree, i'll update to catch a more general exception.

About HTTP_503_SERVICE_UNAVAILABLE, it should still be raised, because download service is unavailable if any subsystem fails.

PR #23, opened 2024-01-30
Harmonize rs-server search with stac search

Padeanu:
In this PR. according to jira 113, to following changes were added:

start_date and end_date are now unified inside a single datetime ({start_date}/{stop_date})
Added a limit argument to display only first N outputs from search.
Search endpoint now returns a FeatureCollection with all information received by eodag from OData response, compatible with STAC.


services/common/rs_server_common/utils/utils.py
```diff
@@ -49,33 +49,47 @@ def is_valid_date_format(date: str) -> bool:
...
+        - The input interval should be in the format "start_date/stop_date"
+        (e.g., "2022-01-01T00:00:00Z/2022-01-02T00:00:00Z").
+        - This function checks for missing start/stop and validates the ISO 8601 format of start and stop dates.
+        - If there is an error, err_code and err_text provide information about the issue.
     """
+
+    try:
+        start_date, stop_date = interval.split("/")
+    except ValueError:
+        return None, None, status.HTTP_422_UNPROCESSABLE_ENTITY, "Missing start/stop"
```
jgaucher-cs, line 87: It would be easier here to just:
```python
from fastapi import HTTPException
raise HTTPException (status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail="Missing start/stop")
```
instead of returning err_code, err_text and return a JSONResponse from the caller.

services/common/rs_server_common/utils/utils.py
```diff
@@ -49,33 +49,47 @@ def is_valid_date_format(date: str) -> bool:
...
     if (not is_valid_date_format(start_date)) or (not is_valid_date_format(stop_date)):
         logger.error("Invalid start/stop in endpoint call!")
-        return False, JSONResponse(
-            status_code=status.HTTP_400_BAD_REQUEST,
-            content="Invalid request, invalid start/stop format",
-        )
-    return True, None
+        return None, None, status.HTTP_400_BAD_REQUEST, "Invalid request, invalid start/stop format"
```
jgaucher-cs, line 90: idem

services/common/rs_server_common/utils/utils.py
```diff
@@ -312,3 +337,54 @@ def eodag_download(argument: EoDAGDownloadHandler, db, init_provider: Callable[[
...
+        stac_template["numberReturned"] += 1
+        stac_template["features"].append(feature_tmp)
+    return stac_template
+
+
+def sort_feature_collection(feature_collection: dict, sortby: str = "+datetime") -> dict:
+    """This function sorts a STAC feature collection based on a given criteria"""
+    # Force default sorting even if the input is invalid, don't block the return collection because of sorting.
+    order = "+" if sortby[0] not in ["+", "-"] else sortby[0]
+    if len(feature_collection["features"]) and hasattr(feature_collection["features"][0], "properties"):
```
jgaucher-cs, line 383: Instead of hasattr(feature_collection["features"][0], "properties")
I think it should be "properties" in feature_collection["features"][0]

PR #24, opened 2024-02-02
Fix/datetime format

jgaucher-cs:


agrosu1978:
Adding the filelock strategy seems to solve the issue when both rs-server-adgs and rs-server-cadip  want to create the same table. I tested, not more errors. For me, this can be merged

agrosu1978 (approved):
Tested the filelock logic addition

Padeanu (approved):

PR #38, opened 2024-02-29
Feat rspy139 - Catalog download asset

Padeanu:
This PR generates a url that can download files from s3 using information stored in catalog.

Sending a GET request to "/catalog/toto/collections/S1_L1/items/fe916452-ba6f-4631-9154-c249924a122d/download/COG" will return a 30 minute link (http://localhost:8077/catalog-bucket/S1_L1/images/may24C355000e4102500n.tif?AWSAccessKeyId=testing&Signature=Rtx4Y1j16lgGWMB%2BWSjMo6LI1LA%3D&Expires=1709214912) that can directly download file stored in catalog (asset alternate href).

### DIFF
--- a/services/catalog/rs_server_catalog/user_catalog.py
+++ b/services/catalog/rs_server_catalog/user_catalog.py
@@ -156,6 +156,14 @@ class UserCatalogMiddleware(BaseHTTPMiddleware):
         content.update({"collection": f"{user}_{content['collection']}"})
         return content
 
+    def generate_presigned_url(self, content, path):
+        # Assume that pgstac already selected the correct asset id
+        # just check type, generate and return url
+        # s3_path = content['asset']
+        import pdb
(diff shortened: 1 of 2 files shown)


services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -33,6 +34,7 @@
 from starlette.middleware.base import BaseHTTPMiddleware
 from starlette.responses import JSONResponse
 
+PRESIGNED_URL_EXPIRATION_TIME = 1800  # 30 minutes
```
agrosu1978, line 37: The expiration time interval must be a configuration parameter of RS-Server (30 min by default). So it should be set from the outside, a suggestion would be to use an env var. Later on, I guess it would be added as a parameter in the deploying helm chart.

PR #47, opened 2024-03-21
RSPY 152: Simplify catalog endpoints

Padeanu:
Swagger Docs to be added for each endpoint.


services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -314,72 +316,78 @@ async def manage_search_response(self, request: Request, response: StreamingResp
...
         content = json.loads(b"".join(map(lambda x: x if isinstance(x, bytes) else x.encode(), body)).decode())
         if request.scope["path"] == "/":  # /catalog/owner_id
             return JSONResponse(content, status_code=response.status_code)
         if request.scope["path"] == "/collections":  # /catalog/owner_id/collections
             content["collections"] = filter_collections(content["collections"], user)
             content = self.remove_user_from_objects(content, user, "collections")
-            content = self.adapt_links(content, ids["owner_id"], ids["collection_id"], "collections")
+            content = self.adapt_links(
+                content,
+                self.request_ids["owner_id"],
```
cyouri-csgroup, line 369: maybe we should not pass ids[] to parameters since we don't do it for manage_put_post_request ?

Padeanu: Agree, can be removed if owner_id, collection_id from :297 are the same as the ones from self.request_ids.

PR #56, opened 2024-03-29
Implement the parallelization for eodag

agrosu1978:
Implement the parallelization for eodag, with the help of a new added env var


services/common/rs_server_common/data_retrieval/eodag_provider.py
```diff
@@ -14,17 +18,25 @@ class EodagProvider(Provider):
...
         Args:
             config_file: the path to the eodag configuration file
             provider: the name of the eodag provider
         """
+        self.eodag_cfg_dir = tempfile.TemporaryDirectory()  # pylint: disable=consider-using-with
         self.provider: str = provider
         self.client: EODataAccessGateway = self.init_eodag_client(config_file)
         self.client.set_preferred_provider(self.provider)
 
+    def __del__(self):
```
Padeanu, line 35: This is not needed, since self.eodag_cfg_dir is from tempfile, therefore is automatically removed when out of context. After tests this command fails.
image

tests/test_download_endpoint.py
```diff
@@ -566,3 +575,110 @@ def test_upload_to_s3(
...
+            download_locations.append(filename[1] + f"/{filename[0]}")
+            # Compose endpoint call and create a list of threads
+            request_threads.append(
+                Thread(target=client.get, args=(f"{endpoint}?name={filename[0]}&local={filename[1]}",)),
+            )
+    # Start all threads in parallel
+    for req_thread in request_threads:
+        req_thread.start()
+    # join threads
+    for req_thread in request_threads:
```
Padeanu, line 672: req_thread just send a starlette test_client request, why join() is needed?

tests/test_download_endpoint.py
Padeanu, line 1: Maybe add a fixed delay (global const) in each test where nedeed, not different values in each test because is misleading (0.4, 2, 3)

PR #123, opened 2024-04-12
Feat/apikey by query

jgaucher-cs:


services/catalog/rs_server_catalog/landing_page.py
```diff
@@ -25,13 +25,18 @@ def add_catalogs(request: Request, auth_roles: list, user_login: str, content: d
         if match := re.match(catalog_read_right_pattern, role):
             groups = match.groupdict()
             if user_login != groups["owner_id"] and groups["collection_id"] == "*" and groups["right_type"] == "read":
+
+                urls = str(request.url).split("?")  # split by query params
+                href = f"{urls[0]}{groups['owner_id']}"
+
+                # To be discussed: maybe we should add the query params (urls[1:])
```
jgaucher-cs, line 32: @cyouri-csgroup we should see that together

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -428,8 +427,8 @@ async def manage_get_response(
         if "detail" not in content:  # Test if the user is authenticated.
             if request.scope["path"] == "/":  # /catalog
                 try:
-                    api_key = request.headers["x-api-key"]
-                    auth_roles, _, user_login = await apikey_security(request, api_key)
+                    auth_roles = request.state.auth_roles
```
jgaucher-cs, line 430: @cyouri-csgroup the authentication info is already in the request.state. We must not call apikey_security manually.

services/catalog/tests/test_authentication_catalog.py
```diff
@@ -241,4 +226,145 @@ async def sub_authentication(client, monkeypatch, httpx_mock: HTTPXMock):
...
+    for pass_the_apikey in PASS_THE_APIKEY:
+        all_collections = client.request("GET", "/catalog/collections", **pass_the_apikey)
+
+        assert all_collections.status_code == HTTP_200_OK
+        content = json.loads(all_collections.content)
+        assert content["collections"] == valid_collections
+
+
+"""
+TODO for Youri, fix these tests
```
jgaucher-cs, line 239: @cyouri-csgroup I let you fix these tests. Also, you should implement tests that verify that we get an error status code if we pass no apikey or an invalid apikey (as in test_authentication.py)

services/common/rs_server_common/authentication.py
```diff
@@ -134,19 +157,31 @@ def apikey_validator(station, access_type):
     def decorator(func):
         @wraps(func)
         def wrapper(*args, **kwargs):
-            if settings.cluster_mode():
+            if settings.CLUSTER_MODE:
```
jgaucher-cs, line 160: @agrosu1978 here I've modified your code to implement a specific error message when the cadip station is unknown, let me know if it's ok for you.

services/common/rs_server_common/db/database.py
```diff
@@ -64,7 +64,7 @@ def open_session(self, url: str = ""):
         # make sure to initialize the session only once.
         with DatabaseSessionManager.lock:
             if (self._engine is None) or (self._sessionmaker is None):
-                self._engine = create_engine(url or self.url(), poolclass=NullPool)
+                self._engine = create_engine(url or self.url(), poolclass=NullPool, pool_pre_ping=True)
```
jgaucher-cs, line 67: Seen with Nicolas: without that the database sort of goes to sleep when unused and we get an error message when we "wake it up" for the first time

PR #144, opened 2024-04-22
Feat rspy162/otlp common log trace

jgaucher-cs:


.gitmodules
```diff
@@ -1,3 +0,0 @@
-[submodule "rs-client-libraries"]
```
jgaucher-cs, line 1: I've removed the git submodule because rs-server doesn't use rs-client. I've added some code that is duplicated between rs-server and rs-client but we'll decide later with Vincent if we put it somewhere in common.

services/catalog/rs_server_catalog/main.py
```diff
@@ -210,6 +211,8 @@ async def dispatch(self, request: Request, call_next: Callable):
 app = api.app
 app.openapi = extract_openapi_specification
 
+# Configure OpenTelemetry
+opentelemetry.init_traces(app, "rs.server.catalog")
```
jgaucher-cs, line 215: That's all you need to do to see the span traces in tempo. No need to manually add traces to your functions or endpoints.

cyouri-csgroup (approved):

agrosu1978 (approved):

PR #294, opened 2024-06-14
Feat rspy153/implicit collection owner

agrosu1978:
Implementation of the optional ownerId for all catalog endpoints


services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -604,10 +604,8 @@ async def manage_get_response(  # pylint: disable=too-many-locals, too-many-bran
         ):
             detail = {"error": "Unauthorized access."}
             return JSONResponse(content=detail, status_code=HTTP_401_UNAUTHORIZED)
-        elif "/queryables" in request.scope["path"]:
-            content["$id"] = request.url._url  # pylint: disable=protected-access
         elif (
-            "/collections" in request.scope["path"] and "items" not in request.scope["path"]
+            "/collection" in request.scope["path"] and "items" not in request.scope["path"]
```
Padeanu, line 608: Typo check:
previously it was collection**s**

services/catalog/tests/test_authentication_catalog.py
Padeanu, line 1: Missing test case where 2 collections from 2 different users have the same name (Toto:S1_L1, pyteam:S1_L1) and test that GET /collections with apikey actually selects the correct collections.

Padeanu: Also permission should allow both users to read both collection in order to make test meaningfull.

Padeanu:
https://github.com/RS-PYTHON/rs-server/blob/develop/docs/doc/users/catalog.md should be updated aswell

jgaucher-cs (changes requested):
Should the `owner_id` become optional in all the catalog endpoints ?

image

cyouri-csgroup (approved):

services/catalog/rs_server_catalog/user_catalog.py
agrosu1978: Solved

PR #298, opened 2024-06-17
Feat RSPY-192: search by session

Padeanu:


Padeanu:
To be investigated:

why "session_id": "S2B_20231117033237234567,S1A_20220715090550123456", works correctly but "session_id": "S2B_20231117033237234567, S1A_20220715090550123456", produce a random result?

PR #300, opened 2024-06-18
Feat rspy141/catalog timestamps

cyouri-csgroup:


services/catalog/rs_server_catalog/timestamps_extension.py
```diff
@@ -0,0 +1,66 @@
...
+    Returns:
+        dict: The updated item.
+    """
+    current_time = datetime.datetime.now()
+    item["properties"]["updated"] = current_time.strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+    if operation == "insertion":  # We insert a new item so we create "expires" field for the first time.
+        if expiration:
+            item["properties"]["expires"] = expiration.strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+        else:
+            plus_30_days = current_time + datetime.timedelta(days=30)
```
Padeanu, line 47: Nope, as per JIRA ticket, this time should be configured. Therefore you should use a env variable.

services/catalog/tests/test_endpoints.py
```diff
@@ -378,6 +385,244 @@ def test_delete_non_empty_collection(
...
+        """Test used to verify that the timestamps extension is correctly set up"""
+        # Create moto server and temp / catalog bucket
+        moto_endpoint = "http://localhost:8077"
+        export_aws_credentials()
+        secrets = {"s3endpoint": moto_endpoint, "accesskey": None, "secretkey": None, "region": ""}
+        # Enable bucket transfer
+        os.environ["RSPY_LOCAL_CATALOG_MODE"] = "0"
+        server = ThreadedMotoServer(port=8077)
+        server.start()
+        try:
```
Padeanu, line 413: **Try** should have at least one **except** block.

services/catalog/tests/test_endpoints.py
```diff
@@ -378,6 +385,244 @@ def test_delete_non_empty_collection(
...
+            current_time = datetime.now().date()
+
+            # Test that published field is correctly added
+            assert "published" in feature_data["properties"]
+            published_datetime_format = datetime.strptime(
+                feature_data["properties"]["published"],
+                "%Y-%m-%dT%H:%M:%S.%fZ",
+            )
+            published_date_only = published_datetime_format.date()
+            assert published_date_only == current_time
```
Padeanu, line 448: Maybe you need to check if these 2 values are close, rather than equal. Since send the push item, then you record time, they cannot have the exact same time.

vprivat-ads (approved):

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -759,6 +774,26 @@ def manage_delete_request(self, request: Request):
...
+        """
+
+        response = requests.get(
+            url=request._url._url,  # pylint: disable=protected-access
+            headers=request.headers,
+            timeout=10,
+        )
+
+        content = json.loads(response.content)
+        return (content["properties"]["published"], content["properties"]["expires"])
```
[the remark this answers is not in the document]
vprivat-ads: it should look like this

```python
item = self.client.get_item(item_id, collection_id, request)
return (item.properties["published"], item.properties["expires"])
```

PR #326, opened 2024-07-01
feat: remove docs/doc/api/rest

jgaucher-cs:


Padeanu:
Ok if the removed code doesn't affect the swagger hosted at: https://dev-rspy.esa-copernicus.eu/docs#

agrosu1978 (approved):

PR #334, opened 2024-07-04
User cannot create/delete a collection owned by another user

agrosu1978:


services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -448,6 +452,16 @@ async def manage_put_post_request(self, request: Request) -> Request | JSONRespo
...
+                # Manage a collection creation. The apikey user (or local user if in local mode)
+                # should be the same as the owner field in the body request. In other words, the
+                # apikey user cannot create a collection owned by another user
+                if user != user_login:
+                    detail = {
+                        "error": f"The '{user_login}' user cannot create a \
+collection owned by the '{user}' user",
+                    }
+                    logger.error(detail["error"])
+                    return JSONResponse(content=detail, status_code=HTTP_400_BAD_REQUEST)
```
Padeanu, line 464: Status code should be one of 401, 403, **405**.

Padeanu:
Do we have a test where a valid collection is created, then with a PATCH request the owner is changed?

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -778,6 +794,16 @@ def manage_delete_request(self, request: Request):
             )
         ):
             return False
+        # we don't care for local mode, any user may create / delete collection owned by another user
+        if common_settings.CLUSTER_MODE and self.request_ids["owner_id"] != user_login:
+            # Manage a collection creation. The apikey user (or local user if in local mode)
```
jgaucher-cs, line 799: deletion, not creation

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -778,6 +794,16 @@ def manage_delete_request(self, request: Request):
...
+            # should be the same as the owner field in the body request. In other words, the
+            # apikey user can't create a collection owned by another user
```
jgaucher-cs, line 801: idem, delete not create

agrosu1978:
> Do we have a test where a valid collection is created, then with a PATCH request the owner is changed?

No, because the PATCH method is not yet implemented

services/catalog/rs_server_catalog/user_catalog.py
agrosu1978: Corrected, thanks !

services/catalog/rs_server_catalog/user_catalog.py
agrosu1978: Corrected

services/catalog/rs_server_catalog/user_catalog.py
agrosu1978: Corrected

agrosu1978:
> Do we have a test where a valid collection is created, then with a PATCH request the owner is changed?

No, because the PATCH method is not yet implemented

PR #384, opened 2024-07-22
RSPY321 - Staging service skeleton

Padeanu:


services/staging/rs_server_staging/main.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+# Endpoint to execute the staging process and generate a job ID
+@app.post("/processes/{resource}/execution")
+async def execute_process(request: dict, resource: str):
+    """Used to execute processing jobs."""
+    job_id = str(uuid.uuid4())  # Generate a unique job ID
+    parameters = request.get("parameters", {})
+    if resource not in api.config["resources"]:
+        raise HTTPException(status_code=404, detail=f"Process resource '{resource}' not found")
```
cyouri-csgroup: HTTP_404_NOT_FOUND instead of 404

services/staging/rs_server_staging/main.py
```diff
@@ -0,0 +1,152 @@
...
+        processor = processors[processor_name]
+        result = processor(parameters)
+
+        # Store job status in TinyDB
+        jobs_table.insert({"job_id": job_id, "status": "completedOK"})
+
+        # Process result as needed and return a response
+        return {"job_id": job_id, "message": "Process executed successfully", "result": result}
+
+    raise HTTPException(status_code=404, detail=f"Processor '{processor_name}' not found")
```
cyouri-csgroup: HTTP_404_NOT_FOUND instead of 404

services/staging/tests/test_staging.py
```diff
@@ -0,0 +1,49 @@
...
+"""Test staging module."""
+
+import pytest
+
+
+@pytest.mark.unit
+def test_ping(staging_client):
+    """Test status."""
+    response = staging_client.get("/_mgmt/ping")
+    assert response.status_code == 200
```
cyouri-csgroup: HTTP_200_OK instead of 200

services/staging/rs_server_staging/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# Copyright 2024 CS Group
```
jgaucher-cs: What it the purpose of this file ? It is empty except for the header.

services/staging/rs_server_staging/main.py
```diff
@@ -0,0 +1,152 @@
...
+from pygeoapi.api import API
+from pygeoapi.config import get_config
+from starlette.exceptions import HTTPException as StarletteHTTPException
+from starlette.middleware.cors import CORSMiddleware
+from starlette.requests import Request
+from starlette.responses import JSONResponse
+from starlette.status import HTTP_200_OK, HTTP_404_NOT_FOUND
+from tinydb import Query, TinyDB
+
+from .processors import processors
```
jgaucher-cs: It's best to use absolute imports i.e. `from rs_server_staging.processors import processors`

services/staging/rs_server_staging/main.py
```diff
@@ -0,0 +1,152 @@
...
+    return JSONResponse(status_code=HTTP_200_OK, content="Check")
+
+
+@app.delete("/jobs/{job_id}")
+async def delete_job(job_id):
+    """Should delete a specific job from db."""
+    return JSONResponse(status_code=HTTP_200_OK, content=job_id)
+
+
+@app.get("/jobs/{job_id}/result")
```
jgaucher-cs: From the Jira https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-321 I think it should be `results` (with a trailing 's') not `result` ?

services/staging/tests/conftest.py
```diff
@@ -0,0 +1,42 @@
...
+
+# pylint: disable=C0413
+# flake8: noqa: F402
+
+import os
+import os.path as osp
+from pathlib import Path
+
+# set pygeoapi env variables
+geoapi_cfg = Path(osp.realpath(osp.dirname(__file__))) / "test_config.yml"
```
jgaucher-cs: Is the "test_config.yml" file necessary ? It is the same as `../rs_server_staging/config/config.yml` maybe you could use this one instead ?

If you want to keep test_config.yml it would be best to move it into a /resources folder so the yml an py files are not mixed in the same directory.

services/staging/tests/test_staging.py
```diff
@@ -0,0 +1,50 @@
...
+import pytest
+from starlette.status import HTTP_200_OK
+
+
+@pytest.mark.unit
+def test_ping(staging_client):
+    """Test status."""
+    response = staging_client.get("/_mgmt/ping")
+    assert response.status_code == HTTP_200_OK
+    assert response.json() == {"status": "healthy"}
```
jgaucher-cs: You have an error in your pytests, it's a detail it's just the response of this health endpoint that is different from the other services. Your staging service returns the string "Healthy" while the other services return the json {"status": "healthy"}. I guess it would be best if you also return this json so the configuration will be the same in kubernetes.

jgaucher-cs:
Should we update rs-server-frontend ?

services/staging/rs_server_staging/Dockerfile
Padeanu: Not used, will be removed.

services/staging/rs_server_staging/main.py
Padeanu: Yes

services/staging/rs_server_staging/main.py
Padeanu: Fixed

services/staging/tests/conftest.py
Padeanu: config.yaml it's a configurable deployment file.

It will be moved to resources.

services/staging/tests/test_staging.py
Padeanu: Fixed, liveness probe should only be checked if return is 200OK.

PR #397, opened 2024-07-26
Feat rspy356/search endpoint

cyouri-csgroup:


services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -325,7 +326,20 @@ def find_owner_id(self, ecql_ast: Node) -> str:
...
+        """Check if the collection exists.
+
+        Returns:
+            bool: True if the collection exists, False otherwise
+        """
+        try:
+            await self.client.get_collection(collection_id, request)
+            return True
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            print(f" collection not found: {e}")
```
Padeanu: Don't use print, use logger.

Also, why don't you use a return based on response status code?

*enclosed with a try-except*
return self.client.get_collection(collection_id, request).status_code == 200

services/catalog/tests/test_endpoints.py
```diff
@@ -148,6 +148,29 @@ def test_post_search_in_toto_s1_l1_collection(self, client):  # pylint: disable=
         content = json.loads(response.content)
         assert len(content["features"]) == 2
 
+    def test_search_in_unexisting_collection(self, client):
```
Padeanu, line 151: One question here, you are searching in collection that doesn't exist using a owner_id that **doesn't exist** "tata".

Shouldn't you test with an owner_id that exists "toto" in a collection that doesn't exist?

Padeanu: Or even have both tests.

services/catalog/rs_server_catalog/user_catalog.py
cyouri-csgroup: Because it is not an http request

services/catalog/tests/test_endpoints.py
cyouri-csgroup: I will add these tests thanks :+1:

jgaucher-cs (approved):

PR #441, opened 2024-08-21
RSPY 322: CADIP search STAC view 1/3

Padeanu:


tests/resources/endpoints/cadip_search.yaml
```diff
@@ -0,0 +1,164 @@
...
+      url: http://127.0.0.1:8000
+      secret:
+        username: test
+        password: test
+collections:
+  - id: cadip_session_by_id_list
+    station: cadip
+    query:
+      id: S1A_20170501121534062343,S1A_20240328185208053186
+      limit: 20
```
don-vip: in this configuration we must be able to configure the OData queries performed towards CADIP stations, so it's not 'limit' but 'top'

tests/resources/endpoints/cadip_search.yaml
```diff
@@ -0,0 +1,164 @@
...
+    station: cadip
+    query:
+      satellite: S1A, S2B
+      limit: 20
+
+  - id: cadip_session_by_start_stop_platform
+    station: cadip
+    query:
+      start_date: "2020-02-16T12:00:00Z"
+      stop_date: "2023-02-16T12:00:00Z"
```
don-vip: same here, it's not start_date/stop_date but PublicationDate. Maybe we could use this syntax?
```
PublicationDate: "2020-02-16T12:00:00Z;2023-02-16T12:00:00Z"
```
which should be translated as
```
PublicationDate gt 2020-02-16T12:00:00Z and PublicationDate lt 2023-02-16T12:00:00Z
```

tests/resources/endpoints/cadip_search.yaml
```diff
@@ -0,0 +1,164 @@
...
+    station: nsg
+    query:
+      limit: 100
+  - id: sgs
+    station: sgs
+    query:
+      limit: 100
+  - id: s1_ins
+    station: ins
+    query: $filter=Satellite in ('S1A', 'S1C')
```
don-vip, line 142: Either we keep my original proposition of including the whole OData $filter syntax in the configuration file, or yours of splitting it key by key, but we can't keep both syntaxes.

services/cadip/rs_server_cadip/cadip_utils.py
```diff
@@ -18,12 +18,50 @@
...
+import yaml
 
 DEFAULT_GEOM = {"geometry": "POLYGON((180 -90, 180 90, -180 90, -180 -90, 180 -90))"}
+CADIP_CONFIG = Path(osp.realpath(osp.dirname(__file__))).parent / "config"
+search_yaml = CADIP_CONFIG / "cadip_search_config.yaml"
+
+
+def validate_cadip_config(fp):
+    """Function to validate yaml template, tba. Should we check this?"""
+    accepted_stations = ["cadip", "ins", "mts"]  # pylint: disable=unused-variable # noqa
```
don-vip: please don't hardcode a list of stations in the source code. One goal of the configuration file is exactly to avoid to change the software if we want to add new stations.

services/cadip/rs_server_cadip/cadip_utils.py
```diff
@@ -18,12 +18,50 @@
...
+    accepted_queries = [  # pylint: disable=unused-variable # noqa
+        "id",
+        "platform",
+        "datetime",
+        "start_date",
+        "stop_date",
+        "limit",
+        "sortby",
+    ]
```
don-vip: The list shall match OData parameter names (Satellite, PublicationDate, top, etc.)

don-vip (changes requested):

services/cadip/rs_server_cadip/cadip_utils.py
Padeanu: Agree, this function will be removed, it's not used.

services/cadip/rs_server_cadip/cadip_utils.py
Padeanu: Agree.

tests/resources/endpoints/cadip_search.yaml
Padeanu: Will be updated.

tests/resources/endpoints/cadip_search.yaml
Padeanu: Agree

tests/resources/endpoints/cadip_search.yaml
Padeanu: Ok. will change.

tests/resources/endpoints/cadip_search.yaml
Padeanu: The split of query key by key is because the values and keys are passed to eodag, which build the query itself. We can't directly forward the $filter={...} without a lot of code changes.

PR #442, opened 2024-08-22
Feat rspy346/oauth2

jgaucher-cs:


services/catalog/rs_server_catalog/main.py
```diff
@@ -339,15 +356,40 @@ async def dispatch(self, request, call_next):
...
 
+# In cluster mode, add the oauth2 authentication
+if common_settings.CLUSTER_MODE:
+
+    # Existing middlewares
+    middleware_names = [middleware.cls.__name__ for middleware in app.user_middleware]
+
+    # Insert the SessionMiddleware (to save cookies) after the DontRaiseExceptions middleware.
+    # Code copy/pasted from app.add_middleware(SessionMiddleware, secret_key=cookie_secret)
+    if app.middleware_stack is not None:
```
Padeanu: just
`if app.middleware_stack:`

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -1122,8 +1124,9 @@ async def dispatch(self, request, call_next):  # pylint: disable=too-many-branch
 
         response = await call_next(request)
 
-        # Don't forward responses that fail
-        if response.status_code != 200:
+        # Don't forward responses that fail.
+        # NOTE: the 30x (redirect responses) are use by the oauth2 authentication.
+        if response.status_code not in (HTTP_200_OK, HTTP_302_FOUND, HTTP_307_TEMPORARY_REDIRECT):
             if response is None:
                 return None
```
Padeanu, line 1135: Non related to this PR, should dispatch return other types than JSONResponse / HTTPException?

services/common/rs_server_common/authentication/apikey.py
```diff
@@ -0,0 +1,147 @@
...
+        json = response.json()
+        if "detail" in json:
+            detail = json["detail"]
+        else:
+            detail = json["error"]
+
+    # If this fail, get the full response content
+    except Exception:  # pylint: disable=broad-exception-caught
+        detail = response.read().decode("utf-8")
+
```
Padeanu:
```python
try:
    json = response.json()
    detail = json.get("detail") or json.get("error")
except (ValueError, KeyError):
    detail = response.content.decode("utf-8", errors="ignore")
```

services/common/rs_server_common/authentication/oauth2.py
```diff
@@ -0,0 +1,314 @@
...
+    """Return True if the request comes from a browser, False if from a console (curl, python console, ...)"""
+
+    # See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent
+    try:
+        user_agent = request.headers["user-agent"]
+    except KeyError:
+        return False
+    return any(
+        browser in user_agent
+        for browser in (
```
Padeanu, line 75: We have to define all possible browsers here? Doesn't look maintainable.

services/common/rs_server_common/authentication/oauth2.py
```diff
@@ -0,0 +1,314 @@
...
+            "OPR/",
+            "Opera/",
+            "Presto/",
+            "Edg/",
+        )
+    )
+
+
+async def is_logged_in(request: Request) -> bool:
+    """We know that the user is logged in if the session cookie exists."""
```
Padeanu: Use non personal expressions:  """The user’s login status is confirmed by the presence of a session cookie."""

services/catalog/tests/conftest.py
```diff
@@ -17,22 +17,36 @@
 import os
 import os.path as osp
 import subprocess  # nosec ignore security issue
+from importlib import reload
 
-# We are in local mode (no cluster).
+RSPY_UAC_HOMEPAGE = "http://RSPY_UAC_HOMEPAGE"
```
Padeanu, line 22: Might be problematic for sonarqube

cyouri-csgroup (approved):

services/catalog/rs_server_catalog/user_catalog.py
jgaucher-cs: I don't know, but in my case I'm returning a RedirectResponse and it's working well

services/common/rs_server_common/authentication/oauth2.py
jgaucher-cs: I've looked at other solutions to determine if we're trying to log from a browser or console, and this one seems to be the more secure. I don't think we'll need to update this list because all existing browsers are already in the list, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent#browser_compatibility

services/catalog/tests/conftest.py
jgaucher-cs: Why that ?

PR #484, opened 2024-09-05
Feat rspy197/update to stac fastapi 3.0.0

cyouri-csgroup:
**Description:**
The goal of this PR is to update to the latest version of stac-fastapi 3.0.0.

- Reworks middlewares signature to solve the breaking changes ([#442 - allow user to pass middleware options](https://github.com/stac-utils/stac-fastapi/pull/442)).
- Adapts the new link "queryable".
- Reworks tests: status_code output (200 --> 201 when posting a new collection/item) and new queryable link.


Padeanu (approved):

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -1123,7 +1123,7 @@ async def dispatch(self, request, call_next):  # pylint: disable=too-many-branch
         response = await call_next(request)
 
         # Don't forward responses that fail
-        if response.status_code != 200:
+        if response.status_code not in [200, 201]:
```
jgaucher-cs: Could you use `if not response.is_success` instead ? @Padeanu what do you think ? 
```python
    @property
    def is_success(self) -> bool:
        """
        A property which is `True` for 2xx status codes, `False` otherwise.
        """
        return codes.is_success(self.status_code)
```

services/catalog/rs_server_catalog/user_handler.py
```diff
@@ -180,29 +180,34 @@ def add_user_prefix(  # pylint: disable=too-many-return-statements
     Returns:
         str: The RS server frontend endpoint.
     """
+    new_path = path
+
     if path == "/collections":
-        return CATALOG_COLLECTION
+        new_path = CATALOG_COLLECTION
```
jgaucher-cs, line 192: Why do you use a `new_path` variable, why can't you just return as before ?

services/catalog/rs_server_catalog/user_catalog.py
Padeanu: Yes, i would be nice

services/catalog/rs_server_catalog/user_handler.py
cyouri-csgroup: It's for the code readability + easier to manipulate in case of debugging.

PR #663, opened 2024-10-25
Fix rspy453/fix rs server for stac browser

jgaucher-cs:


services/adgs/rs_server_adgs/adgs_utils.py
```diff
@@ -0,0 +1,141 @@
...
+
+import yaml
+from fastapi import HTTPException, status
+from rs_server_common.stac_api_common import QueryableField, map_stac_platform
+
+ADGS_CONFIG = Path(osp.realpath(osp.dirname(__file__))).parent / "config"
+search_yaml = ADGS_CONFIG / "adgs_search_config.yaml"
+
+
+@lru_cache(maxsize=1)
```
Padeanu: this has to be removed, it's buggy.

Padeanu: Same for cadip

services/adgs/rs_server_adgs/adgs_utils.py
```diff
@@ -0,0 +1,141 @@
...
+    stac_mapper_path = ADGS_CONFIG / "adgs_stac_mapper.json"
+    with open(stac_mapper_path, encoding="utf-8") as stac_map:
+        stac_mapper = json.loads(stac_map.read())
+        return {stac_mapper.get(stac_key, stac_key): value for stac_key, value in stac_params.items()}
+
+
+def serialize_adgs_asset(feature_collection, request):
+    """Used to update adgs asset with propper href and format {asset_name: asset_body}."""
+    for feature in feature_collection.features:
+        feature.assets["file"].href = f"{request.url.scheme}://{request.url.netloc}/adgs/aux?name={feature.id}"
```
Padeanu, line 60: ToDo, update this with staging href

services/common/rs_server_common/stac_api_common.py
```diff
@@ -0,0 +1,619 @@
...
+
+    # adgs or cadip function
+    all_collections: Callable = None
+    select_config: Callable = None
+    stac_to_odata: Callable = None
+    map_mission: Callable = None
+
+    # Is the service adgs or cadip ?
+    adgs: bool = False
+    cadip: bool = False
```
Padeanu, line 92: Maybe we need to change this in future, since we have to add new services, and can't create a variable for each

services/common/rs_server_common/stac_api_common.py
```diff
@@ -0,0 +1,619 @@
...
+                        enum=connstellations,
+                    ),
+                },
+            )
+
+        return queryables
+
+    async def fetchval(self, query, *args, column=0, timeout=None):
+        """Run a query and return a value in the first row.
+
```
Padeanu, line 185: use google style docstr

services/common/rs_server_common/stac_api_common.py
```diff
@@ -0,0 +1,619 @@
...
+                 None if no records were returned by the query.
+        """
+        query = query.strip()
+
+        # From stac_fastapi.pgstac.core.CoreCrudClient::all_collections
+        if query == "SELECT * FROM all_collections();":
+            return filter_allowed_collections(self.all_collections(), self.service, self.request)
+
+        # From stac_fastapi.pgstac.core.CoreCrudClient::get_collection
+        if query == "SELECT * FROM get_collection($1::text);":
```
Padeanu, line 203: This looks like a limitation of pgstac (or whatever we use), we either use ORM or SQL, mixing them it's a confusion factor

services/common/rs_server_common/stac_api_common.py
jgaucher-cs: It's just an easy way to implement specific code like this: 
```python
        if self.adgs:
            stac_params["constellation"], stac_params["platform"] = mission
        if self.cadip:
            stac_params["platform"] = mission
```
of course we'll have to evolve this for future services.

services/common/rs_server_common/stac_api_common.py
jgaucher-cs: This code is taken from `stac_fastapi.pgstac.core.CoreCrudClient::get_collection`, in fact my class has to mock these queries made by stac_fastapi so I cannot change this query.

PR #676, opened 2024-10-29
ADGS adaptation for staging

Padeanu:


services/adgs/config/adgs_ws_config.yaml
```diff
@@ -121,3 +121,104 @@ adgs:
     metadata_mapping:
         uid: '$.Id'
         downloadLink: 'http://127.0.0.1:5000/Products({uid})/$value'
+
+
+
+adgs2:
```
jgaucher-cs, line 126: Is it normal that the fields are so different between the two adgs stations ? 
```yaml
adgs:
  auth:
    type: TokenAuth
    auth_uri: 'http://127.0.0.1:5000/oauth2/token'
    token_type: json
    token_key: access_token
    request_method: POST
```
```yaml
adgs2:
  auth:
    type: HTTPHeaderAuth
    headers:
        Authorization: "Bearer {token}"
    credentials:
      token: token_value
```

Padeanu: No, wrong copy-paste, willfix

PR #677, opened 2024-10-30
Add the service notion when staging processor is created

agrosu1978:


services/staging/rs_server_staging/processors.py
```diff
@@ -260,6 +265,12 @@ def __init__(
         self.catalog_collection: str = collection
         self.catalog_item_name: str = item
         self.provider: str = provider
+        if self.provider in ADGS_STATIONS:
```
Padeanu: use a dictionary, to map provider with stations. also for adgs there is an additional station, adgs2.

```python
table = {"cadip": ['ins', 'mps'], "auxin": ['adgs', 'adgs2']}

self.service = next(key for key, values in table.items() if self.provider in values)
```

Padeanu: Also, hardcoding station here, doesn't look good. Is there any function that reads all the configurations we've written in rs-common that describe stations and return them as a list?

agrosu1978: Solved, see line 339 in services/common/rs_server_common/authentication/authentication_to_external.py

PR #678, opened 2024-11-04
Feat rspy475/add pytests

agrosu1978:
- Add pytests for us 475
- Try an implementation for pagination in build_fileslist_to_be_deleted function from user_catalog. It seems the pgstac is not returning a token for the next page


services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -425,8 +401,6 @@ def update_stac_item_publication(  # pylint: disable=too-many-locals
                 old_bucket_arr[2] = CATALOG_BUCKET
                 s3_key = "/".join(old_bucket_arr)
                 # Check if the S3 key exists
-                # TBD: The catalog should have nothing to do anymore with the
-                # s3 level. This part has to be removed
```
jgaucher-cs, line 425: Should this part be removed in this story or another future story ? Or it doesn't need to be removed anymore ?

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -1040,37 +1014,55 @@ async def manage_delete_response(self, response: StreamingResponse, user: str) -
         if "deleted collection" in response_content:
             response_content["deleted collection"] = response_content["deleted collection"].removeprefix(f"{user}_")
         # delete the s3 files as well
-        self.delete_s3_files()
+        delete_s3_files(self.s3_files_to_be_deleted)
+        self.s3_files_to_be_deleted.clear()
         return JSONResponse(response_content)
 
-    async def build_filelist_to_be_deleted(self, request):
+    async def build_fileslist_to_be_deleted(self, request):
```
jgaucher-cs: In fact the correct form is "file list" not "files list" :) https://english.stackexchange.com/questions/164714/item-queue-vs-items-queue-files-list-vs-file-list

services/catalog/rs_server_catalog/user_catalog.py
agrosu1978: It will remain in fact

services/catalog/rs_server_catalog/user_catalog.py
agrosu1978: Indeed, i set it right the first time :) I will correct it

PR #679, opened 2024-11-04
Fix/rspy449/fix search endpoints

mcolin16:
Fix catalog search endpoint to allow search using multiple collections + refactor some parts of the script user_catalog.py


services/catalog/tests/test_endpoints.py
```diff
@@ -333,6 +330,152 @@ def test_search_with_collections_and_filter(self, client):  # pylint: disable=mi
...
+                    "args": [
+                        {"op": "=", "args": [{"property": "width"}, 2500]},
+                        {"op": "=", "args": [{"property": "height"}, 2500]},
+                    ],
+                },
+            }
+            response = client.post("/catalog/search", json=test_json)
+        elif method == "GET":
+            test_params = {
+                "collections": "S1_L1,toto_S2_L3",
```
Padeanu, line 367: why one collection is prefixed with user and second is not? Does it work on local env without specific owner id?

services/catalog/tests/test_endpoints.py
```diff
@@ -333,6 +330,152 @@ def test_search_with_collections_and_filter(self, client):  # pylint: disable=mi
...
+                "filter-lang": "cql2-text",
+                "filter": "width=2500",
+                "owner": "toto",
+            }
+            response = client.get("/catalog/search", params=test_params)
+        assert response.status_code == fastapi.status.HTTP_200_OK
+        assert len(json.loads(response.content)["features"]) == 3
+
+        # Use implicit naming mechanism for some collections of the list + specify owner in the filter
```
Padeanu, line 375: You don't specify owner in json as described in comment above

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -227,21 +223,21 @@ async def get_item_from_collection(self, request: Request):
         try:
             item = await self.client.get_item(
                 item_id=self.request_ids["item_id"],
-                collection_id=f"{self.request_ids['owner_id']}_{self.request_ids['collection_id']}",
+                collection_id=f"{self.request_ids['owner_id']}_{self.request_ids['collection_id'][0]}",
```
Padeanu, line 226: Why you access index 0 now? What's stored in other indexes that previous wasn t?

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -535,7 +531,7 @@ async def collection_exists(self, request: Request, collection_id: str) -> bool:
             logger.error("Collection %s not found: %s", collection_id, e)
             return False
 
-    async def manage_search_request(  # pylint: disable=too-many-branches
+    async def manage_search_request(  # pylint: disable=too-many-statements,too-many-branches
```
Padeanu, line 534: We soon need a big refactor to simplify and remove all that "too-many-X"

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -548,71 +544,118 @@ async def manage_search_request(  # pylint: disable=too-many-branches
...
             content = await request.json()
-            if (
-                "filter-lang" not in content and "filter" in content
-            ):  # If not specified, the default value of filter_lang in a post method is cql2-json.
-                filter_lang = {"filter-lang": "cql2-json"}
-                stac_filter = {}
-                stac_filter["filter"] = content.pop("filter")
+            # Add filter-lang option to the content if it doesn't already exist
+            if "filter" in content:
+                filter_lang = (
```
Padeanu:
```python
filter_lang = {"filter-lang": content.get("filter-lang", "cql2-json")}
```

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -548,71 +544,118 @@ async def manage_search_request(  # pylint: disable=too-many-branches
...
+                # Management of priority for the assignation of the owner_id
+                filters = parse_cql2_json(content["filter"])
+                owner_id_filter = self.find_owner_id(filters)
+                if owner_id_filter:
+                    owner_id = owner_id_filter
+                elif "owner" in content:
+                    owner_id = content["owner"]
+                else:
+                    owner_id = get_user(self.request_ids["owner_id"], self.request_ids["user_login"])
+                self.request_ids["owner_id"] = owner_id
```
Padeanu:
```python
# Since you don't use any level related debugs, use this to set priority 
self.request_ids["owner_id"] = (
    self.find_owner_id(filters)
    or content.get("owner")
    or get_user(self.request_ids["owner_id"], self.request_ids["user_login"])
)

```

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -548,71 +544,118 @@ async def manage_search_request(  # pylint: disable=too-many-branches
...
+        elif request.method == "GET":
+            # Get dictionary of query parameters
+            query_params_dict = dict(request.query_params)
+
+            # Update owner_id if it is not already defined from path parameters
+            if not self.request_ids["owner_id"]:
+                owner_id_filter = (
+                    self.find_owner_id(parse_ecql(query_params_dict["filter"])) if "filter" in query_params_dict else ""
+                )
+                if owner_id_filter:
```
Padeanu: same as above, also seems like can be reused into a small function

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -1141,65 +1207,121 @@ async def retrieve_timestamp(self, request: Request) -> tuple[str, str]:
...
-        if request.scope["path"] == "/search":
-            # URL: GET: '/catalog/search'
-            request = await self.manage_search_request(request)
-            if hasattr(request, "status_code"):  # Unauthorized
-                return request
-        elif request.method in {"POST", "PUT"} and self.request_ids["owner_id"]:
+
+        # ---------- Apply specific changes for each endpoint
+
+        if request.method in {"POST", "PUT"} and "/search" not in request.scope["path"]:
```
Padeanu: Not very important, but fyi, it's optimal to use this operator over Tuples, not sets, since are immutables thus lower memory footprint

services/catalog/tests/test_endpoints.py
Padeanu, line 1: Not sure if it's very related to this PR (more related to apikey), but i see tests when you search in group of valid, invalid collection.

Shouldn't we have tests with 2 valid collections, but without read roles over one?

services/catalog/tests/test_endpoints.py
```diff
@@ -333,6 +330,152 @@ def test_search_with_collections_and_filter(self, client):  # pylint: disable=mi
...
+        response = client.get("/catalog/search", params=test_params)
+        assert response.status_code == 200
+        content = json.loads(response.content)
+        assert len(content["features"]) == 3
+
+        # Use implicit naming mechanism for some collections of the list + specify owner in the filter
+        test_params = {
+            "collections": "S1_L1,toto_S2_L3",
+            "filter-lang": "cql2-text",
+            "filter": "width=2500 AND owner='toto'",
```
Padeanu: Why do we need that width?

services/catalog/tests/test_endpoints.py
```diff
@@ -333,6 +330,152 @@ def test_search_with_collections_and_filter(self, client):  # pylint: disable=mi
...
+        }
+        response = client.get("/catalog/search", params=test_params)
+        assert response.status_code == 200
```
Padeanu: Use fastapi.status.HTTP_* as much as possible

services/catalog/tests/test_endpoints.py
```diff
@@ -333,6 +330,152 @@ def test_search_with_collections_and_filter(self, client):  # pylint: disable=mi
...
+        assert len(content["features"]) == 3
+
+        # Implicit naming mechanism will not produce the right owner_id if we don't specify it in the query
+        # parameters or in the filter
+        test_params = {"collections": "toto_S1_L1,S2_L3", "filter": "width=2500"}
+        response = client.get("/catalog/search", params=test_params)
+        assert response.status_code == fastapi.status.HTTP_404_NOT_FOUND
+
+        # Check that we get an error if at least one collection does not exist
+        test_params = {"collections": "toto_S1_L1,unexisting_collection", "filter": "width=2500"}
```
Padeanu: What happens here if the filter contains also the "owner" : "toto"?

services/catalog/rs_server_catalog/search.py
Padeanu, line 1: This code was moved into user_catalog.py?


If yes, we need to divide that file as much as possible, since it's already very big and almost unreadable

services/catalog/tests/test_endpoints.py
mcolin16: It is to show that you can choose to specify or not the owner_id in the name of the collection and both cases will work because you specified the owner_id in the filter
If no owner_id is specified in local-mode (neither in the name of the collection nor in the filter), the default owner id will be selected and concatenated to the name of the collection

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -670,48 +722,54 @@ async def manage_put_post_request(  # pylint: disable=too-many-branches
...
                     self.request_ids["owner_id"],
-                    user_login,
+                    self.request_ids["user_login"],
                 )
             ):
                 detail = {"error": "Unauthorized access."}
                 return JSONResponse(content=detail, status_code=HTTP_401_UNAUTHORIZED)
 
+            if len(self.request_ids["collection_id"]) > 1:
+                detail = {"error": "Cannot create or update more than one collections !"}
```
Padeanu: collection*

services/catalog/tests/test_endpoints.py
mcolin16: Here he owner_id is specified in the body (POST method) or in the query parameters (GET method)

services/catalog/rs_server_catalog/user_catalog.py
mcolin16: This is because the collection_id is now a list and not a string (because we are now able to use several collections for the search endpoint). In some use cases such as the creation of a new item in a collection, there is only one collection specified in the list "collection_id" (because we create an item that will belong to only one collection), that's why just have to get the first (and only) element of the list

services/catalog/tests/test_endpoints.py
mcolin16: There are already such tests in test_authentification_catalog.py (for example the test test_fails_without_good_perms)

services/catalog/rs_server_catalog/search.py
mcolin16: Yes because I think it was better to keep the content of search.py in the user_catalog, but I will try to divide the user_catalog file in a different way in another PR

services/catalog/tests/test_endpoints.py
mcolin16: The code first check if the original name of the collection exist in the database, if it's the case it will keep this name, otherwise the owner_id from the filter will be concatenated to the name of the collection

PR #683, opened 2024-11-06
Fix the type changing from str to str list

agrosu1978:
Fix the changing of the type for self.client.request_ids['collection_id'] from str to a str list

### DIFF
--- a/services/catalog/rs_server_catalog/user_catalog.py
+++ b/services/catalog/rs_server_catalog/user_catalog.py
@@ -1081,60 +1081,63 @@ field is not permitted also.",
 
     async def build_filelist_to_be_deleted(self, request):
         """Build the list of the s3 files that will be deleted if the request is successfull"""
-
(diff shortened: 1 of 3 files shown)


services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -1081,60 +1081,63 @@ async def manage_delete_response(self, response: StreamingResponse, user: str) -
...
-                # NOTE: Unable to use the pagination from pgstac client. Temporary, use a limit of 100
-                token = None
-                while True:
-                    items_collection = await self.client.item_collection(
-                        request=request,
+        for ci in self.request_ids["collection_id"]:
+            collection_id = f"{self.request_ids['owner_id']}_{ci}"
+            items = []
+            try:
+                await self.client.get_collection(collection_id, request)
```
jgaucher-cs: Why do you need this ? You don't use the returned value ?

agrosu1978: i have just forgotten about it, it was left from a test. i shall remove it

PR #688, opened 2024-11-12
Feat/sync config files

jgaucher-cs:
Please ignore whitespaces when comparing changes: https://github.com/RS-PYTHON/rs-server/pull/688/files?w=1

**Instructions:** run `poetry run python resources/sync_config_files.py` from the rs-server root directory to create all configuration files from templates and copy them to rs-demo, rs-helm, rs-infrastructure (these projects must be checked out in your same local directory than rs-server)


resources/sync_config_files.py
```diff
@@ -0,0 +1,705 @@
+# Copyright 2024 CS Group
```
jgaucher-cs, line 1: **Note:** run this script with `poetry run python resources/sync_config_files.py` from the rs-server root directory to create all configuration files from templates and copy them to rs-demo, rs-helm, rs-infrastructure (these projects must be checked out in your same local directory than rs-server)

services/adgs/config/adgs_search_config.template.yaml
```diff
@@ -0,0 +1,68 @@
+# Copyright 2024 CS Group
```
jgaucher-cs, line 1: This is the template file for `adgs_search_config.yaml`. All common values are set only once in a root `template` block.

services/adgs/config/adgs_search_config.template.yaml
```diff
@@ -0,0 +1,68 @@
...
+      roles:
+        - producer
+        - licensor
+      url: "https://sentiwiki.copernicus.eu/web/s1-mission"
+    - name: "Reference System"
+      roles:
+        - host
+      url: "https://home.rs-python.eu/"
+
+collections:
```
jgaucher-cs, line 48: This structure is compliant with the old `adgs_search_config.yaml` file, it contains specific values for each sub-block.

jgaucher-cs, line 48: This structure is compliant with the old `adgs_search_config.yaml` file, it contains specific values for each sub-block.

services/adgs/config/adgs_search_config.yaml
```diff
@@ -12,87 +12,117 @@
...
-      - name: 'European Union/ESA/Copernicus'
-        roles:
-          - producer
-          - licensor
-        url: 'https://sentiwiki.copernicus.eu/web/s1-mission'
-      - name: 'Reference System'
-        roles:
-          - host
-        url: 'https://home.rs-python.eu/'
+# THIS FILE WAS AUTOMATICALLY CREATED FROM:
```
jgaucher-cs, line 15: This file was automatically created from the template when running `poetry run python resources/sync_config_files.py`

services/adgs/config/adgs_search_config.yaml
```diff
@@ -12,87 +12,117 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-collections:
-  - id: s1_adgs_OPER_AUX_ECMWFD_PDMC
-    station: adgs
-    query:
-      productType: OPER_AUX_ECMWFD_PDMC
-    stac_extensions: [ "https://stac-extensions.github.io/eo/v1.0.0/schema.json", "https://stac-extensions.github.io/projection/v1.0.0/schema.json", "https://stac-extensions.github.io/view/v1.0.0/schema.json" ]
```
jgaucher-cs, line 20: Note: python yaml does some reformatting but this doesn't change the way the file is read afterwards.

services/adgs/config/adgs_ws_config.template.yaml
```diff
@@ -0,0 +1,151 @@
+# Copyright 2024 CS Group
```
jgaucher-cs, line 1: Template for `adgs_ws_config.yaml`

services/adgs/config/adgs_ws_config.yaml
```diff
@@ -35,98 +40,96 @@ adgs:
...
+            and att/OData.CSC.StringAttribute/Value eq '{attr_processing_date#to_iso_utc_datetime}')
     pagination:
       max_items_per_page: 1000
       next_page_url_tpl: '{url}?{search}&$top={items_per_page}&$expand=Attributes'
-      total_items_nb_key_path: '$.total'
-
+      total_items_nb_key_path: $.total
   download:
     type: HTTPDownload
-    base_uri: 'http://127.0.0.1:5001/Products'
```
jgaucher-cs, line 121: @agrosu1978 @Padeanu why did we have 5001 here (that corresponds to the adgs2 station) instead of 5000 ? For pytests, or a mistake ?

services/adgs/config/adgs_ws_config_token_module.yaml
```diff
@@ -12,201 +12,204 @@
...
+            and att/OData.CSC.StringAttribute/Value eq '{attr_processing_date#to_iso_utc_datetime}')
     pagination:
       max_items_per_page: 1000
       next_page_url_tpl: '{url}?{search}&$top={items_per_page}&$expand=Attributes'
-      total_items_nb_key_path: '$.total'
-
+      total_items_nb_key_path: $.total
   download:
     type: HTTPDownload
-    base_uri: 'http://127.0.0.1:5001/Products'
```
jgaucher-cs, line 110: @agrosu1978 @Padeanu same here, why did we have 5001 ?

services/cadip/config/cadip_ws_config.template_session.yaml
```diff
@@ -0,0 +1,230 @@
+# Copyright 2024 CS Group
```
jgaucher-cs, line 1: Here we use 2 template files (because the `_session` blocks have different values) that will be merged into the single `cadip_ws_config.yaml` file

services/cadip/config/cadip_ws_config.yaml
```diff
@@ -28,101 +34,92 @@ cadip:
...
     metadata_mapping:
-        uid: '$.Id'
-        downloadLink: 'http://127.0.0.1:5000/Files({uid})/$value'
-
+      uid: $.Id
+      downloadLink: http://127.0.0.1:5000/Files({uid})/$value
 ins:
   auth:
     type: TokenAuth
-    auth_uri: 'http://127.0.0.1:0/oauth2/token'
```
jgaucher-cs, line 116: @agrosu1978 @Padeanu why the port was 0 here ? For pytests or a mistake ?

services/adgs/config/adgs_ws_config.yaml
Padeanu: It's a mistake, every port related to ADGS should be 5000, while every port related to ADGS2 should be 5001.

Make sure that in unit test responses are not mocked with :5001

services/cadip/config/cadip_ws_config.yaml
Padeanu: Typo, should be 5100

services/cadip/config/cadip_ws_config.yaml
```diff
@@ -28,101 +34,92 @@ cadip:
...
+    auth_uri: http://127.0.0.1:5100/oauth2/token
     token_type: json
     token_key: access_token
     request_method: POST
     req_data:
-      client_id: client
-      client_secret: client_sec
+      client_id: client_id
```
jgaucher-cs: @agrosu1978 @Padeanu why do stations have different id and secret ? For pytests ? Or do we want that in production = in the helm charts ?

services/cadip/config/cadip_ws_config.yaml
```diff
@@ -28,101 +34,92 @@ cadip:
...
+      client_secret: client_secret
       grant_type: password
-      username: test2
```
jgaucher-cs, line 124: @agrosu1978 @Padeanu same question here, is it normal to have username and password in req_data AND in credentials ?

Padeanu: @agrosu1978

services/cadip/config/cadip_ws_config.yaml
Padeanu: @agrosu1978

services/cadip/config/cadip_ws_config.yaml
```diff
@@ -132,96 +129,88 @@ ins:
...
     metadata_mapping:
-        uid: '$.Id'
-        downloadLink: 'http://127.0.0.1:5100/Files({uid})/$value'
-
+      uid: $.Id
+      downloadLink: http://127.0.0.1:5100/Files({uid})/$value
 mps:
   auth:
     type: TokenAuth
-    auth_uri: 'http://127.0.0.1:0/oauth2/token'
```
jgaucher-cs, line 215: @agrosu1978 @Padeanu same questions here and in the next stations for the port, id, secret, username and password

services/cadip/config/cadip_ws_config.yaml
```diff
@@ -765,121 +718,116 @@ ins_session:
       collection: Sentinel1
   search:
     type: ODataV4Search
-    api_endpoint: "http://127.0.0.1:5000/Sessions"
```
jgaucher-cs, line 768: @agrosu1978 @Padeanu why the port was 5000 instead of 5100 for ins_session ?

jgaucher-cs, line 768: @agrosu1978 @Padeanu why the port was 5000 instead of 5100 for ins_session ?

services/cadip/config/cadip_ws_config.yaml
Padeanu: Port typo, should be 5101.

Since this file is only applicable in local mode, i'm not sure we really use that information, @agrosu1978  should we update or remove them?

services/cadip/config/cadip_ws_config.yaml
Padeanu: Typo, should be 5100

Padeanu: Same

services/common/config/rs-server.yaml
```diff
@@ -12,75 +12,72 @@
...
+# THIS FILE WAS AUTOMATICALLY CREATED FROM:
+#  - rs-server/services/common/config/rs-server.template.yaml
+#  - rs-server/resources/sync_config_files.py
+# DON'T MODIFY IT DIRECTLY !
+
 external_data_sources:
-  # AUXIP. For now only one ADGS but in the future there will be another one
   adgs:
-    domain: mockup-auxip-adgs-svc.processing.svc.cluster.local
+    domain: mockup-station-adgs-svc.processing.svc.cluster.local
```
jgaucher-cs, line 22: @agrosu1978 @Padeanu is this domain value used somewhere ? Here I've taken the values used in the helm charts to be consistant. I guess this value (that describes a pod url in the cluster) has no meaning in rs-server in local mode.

jgaucher-cs, line 22: @agrosu1978 @Padeanu is this domain value used somewhere ? Here I've taken the values used in the helm charts to be consistant. I guess this value (that describes a pod url in the cluster) has no meaning in rs-server in local mode.

services/common/config/rs-server.yaml
```diff
@@ -12,75 +12,72 @@
...
-      scope:
+      scope: null
       username: test
       password: test
       client_id: client_id
       client_secret: client_secret
       authorization: Basic test
-
-  # CADIP stations. MPS has Authorization but no scope. MTI has scope but no Authorization
-  # the following one is for tests onlys
```
jgaucher-cs, line 50: @agrosu1978 @Padeanu the comment says that this is for tests only... but these values are used in production in the helm charts, is this normal ?

services/cadip/config/cadip_ws_config.yaml
jgaucher-cs: In fact the values from this file will be merged into rs-helm and rs-infrastructure so maybe we need the values here

services/cadip/config/cadip_ws_config.yaml
agrosu1978: These fields are overwritten. You can set whatever you want for the following (in cadip_ws_config.yaml or adgs_ws_config.yaml files)

auth__auth_uri
auth__req_data__client_id
auth__req_data__client_secret
auth__req_data__username
auth__req_data__password
auth__req_data__grant_type
auth__credentials__username
auth__credentials__password

Check the set_eodag_auth_env function from common/rs_server_common/authentication/authentication_to_external.py

Maybe a PLACEHOLDER should be set as a value for all these fields.

services/cadip/config/cadip_ws_config.yaml
agrosu1978: It is used by eodag (when it is configured to handle by itself the token from the external stations), i guess it should not be removed

services/common/config/rs-server.yaml
agrosu1978: I used the configuration provided in the user story 352, please have a look at:
`Default configuration (local mode format. To be converted to Kubernetes secrets for cluster mode):`
The values from the domain are not yet used, but they will probably be:
`one function that returns the configuration for a given URL, by looking at which entry defines a domain that matches the URL domain, or raises an error if not exactly one entry is found`

agrosu1978: Same as [here](https://github.com/RS-PYTHON/rs-server/pull/688#discussion_r1838372601)

services/cadip/config/cadip_ws_config.yaml
jgaucher-cs: @agrosu1978 so can I remove these fields from the config file ? Or do you prefer to keep them ?

services/common/config/rs-server.yaml
agrosu1978: Yes, probably I forgot about that comment with tests only. The `authentication.token_url` field is the important one, because that is the link used to get the token. So it should be configured differently between local_mode and cluster_mode.
On the other hand, at some point, the `cadip_session` like sections will have actually to replace the `cadip` sections everywhere, in rs-server.yaml, cadip/adgs_ws_config.yaml and cadip/adgs_ws_config_token_module.yaml   files

services/adgs/config/adgs_ws_config.yaml
jgaucher-cs: @Padeanu indeed this is why the pytests were not passing, thank you

services/common/config/rs-server.yaml
jgaucher-cs: OK so for now we can use any value since this is not used, I suggest to keep the value (in green above) that is defined in the helm charts for now

services/cadip/config/cadip_ws_config.yaml
agrosu1978: I prefer to keep them, to be the same as in the story

services/cadip/config/cadip_ws_config.yaml
agrosu1978: Stations may have different ids and secrets in real life, of course. It is up to the operational team to set them accordingly

PR #690, opened 2024-11-14
Feat RSPY131: RS Server pagination

Padeanu:


services/cadip/rs_server_cadip/fastapi/cadip_app.py
```diff
@@ -25,7 +25,7 @@
 from rs_server_common.fastapi_app import init_app
 
 # Init the FastAPI application with the cadip routers.
-app = init_app(__version__, cadip_routers, init_db=True, router_prefix="/cadip")
+app = init_app(__version__, cadip_routers, init_db=False, router_prefix="/cadip")
```
jgaucher-cs: Why do you need this ? What does it change ?

Padeanu: Faster debug in local mode, it doesn't require a db running when starting this service.
Can be set back to True, but will be removed in future since we don't use a DB anymore (was used in old versions for status)

services/adgs/rs_server_adgs/api/adgs_search.py
```diff
@@ -302,13 +320,16 @@ def process_product_search(  # pylint: disable=too-many-locals
         HTTPException (fastapi.exceptions): If there is a general failure during the process.
     """
     set_eodag_auth_token(station, "auxip")
-    limit = limit if limit else 1000
+    limit = limit if limit else 1000  # needed?
```
jgaucher-cs: I guess not needed because above you added: 
```python
user_limit = int(self.request.query_params.get("limit", odata_params.get("top", 1000)))
```

services/cadip/rs_server_cadip/api/cadip_search.py
```diff
@@ -85,13 +85,27 @@ def __init__(self, request: Request | None = None, readwrite: Literal["r", "w"]
     @handle_exceptions
     async def process_search(self, collection: dict, odata_params: dict) -> stac_pydantic.ItemCollection:
         """Do the search for the given collection and OData parameters."""
+        sortby = self.request.query_params.get("sortby", None)
+        user_limit = int(self.request.query_params.get("limit", odata_params.get("top", 10)))
```
jgaucher-cs: Can you add the same comment as in adgs ? 
```python
# Priority, stac endpoint limit -> odata collection top -> 1000 by default
```

services/cadip/rs_server_cadip/api/cadip_search.py
```diff
@@ -381,7 +398,7 @@ def process_session_search(  # type: ignore  # pylint: disable=too-many-argument
         HTTPException (fastapi.exceptions): If there is a value error during mapping.
     """
     limit = limit if limit else 1000
```
jgaucher-cs: As in adgs, maybe this is not needed because above you added:
```python
user_limit = int(self.request.query_params.get("limit", odata_params.get("top", 10)))
```

services/adgs/rs_server_adgs/api/adgs_search.py
```diff
@@ -83,11 +84,25 @@ def __init__(self, request: Request | None = None, readwrite: Literal["r", "w"]
...
+            except (ValueError, TypeError) as exc:
+                # Handle cases where `page` is not an integer or is invalid
+                raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="Page number invalid") from exc
         return process_product_search(
             collection.get("station", "adgs"),
             odata_params.get("productType"),
             odata_params.get("PublicationDate"),
-            odata_params.get("top", None),
+            user_limit,
+            self.request.query_params.get("sortby", "-created"),
```
jgaucher-cs, line 93: Is it necessary to set default value "-created", because in the process_product_search function below, this is already the default value. It would be better to use this default value at a single place to improve maintainability.

e.g. it could take the default value "-created" here, and be a mandatory argument in the process_product_search function below

services/cadip/rs_server_cadip/api/cadip_search.py
```diff
@@ -85,13 +85,27 @@ def __init__(self, request: Request | None = None, readwrite: Literal["r", "w"]
     @handle_exceptions
     async def process_search(self, collection: dict, odata_params: dict) -> stac_pydantic.ItemCollection:
         """Do the search for the given collection and OData parameters."""
+        sortby = self.request.query_params.get("sortby", None)
```
jgaucher-cs: if "sortby" is missing from the arguments, it will take None, and its value will be None also in the process_session_search function below. As for adgs, maybe it would be better if it took the default value "-datetime" here, and be a mandatory argument in the process_session_search function below.

services/adgs/rs_server_adgs/api/adgs_search.py
Padeanu: Ok removed from signature

services/cadip/rs_server_cadip/api/cadip_search.py
Padeanu: Removed from signature, default is now - published (PublicationDate desc)

services/common/rs_server_common/stac_api_common.py
```diff
@@ -503,12 +543,40 @@ def read_cql(filt: dict):
...
+        return data
 
     @abstractmethod
     async def process_search(self, collection: dict, odata_params: dict) -> stac_pydantic.ItemCollection:
         """Do the search for the given collection and OData parameters."""
 
+    def get_page(self) -> int:
+        """Get page value (integer) according to given priority"""
+        try:
+            user_page = int(self.request.query_params.get("page", self.odata.get("page", 1)))  # type: ignore
```
jgaucher-cs: This doesn't work, `self.odata.get("page")` doesn't exit. You should use `self.page` instead.

Padeanu: Agree

PR #692, opened 2024-11-15
RSPY-515 - Use Origin instead of Referer to identify STAC Browser

vprivat-ads:


vprivat-ads:
@jgaucher-cs tested OK with real station on our infra

PR #705, opened 2024-11-22
Feat rspy342/persist staging jobs julien

jgaucher-cs:


services/common/rs_server_common/utils/utils2.py
```diff
@@ -47,3 +52,28 @@ def read_response_error(response):
         detail = response.content.decode("utf-8", errors="ignore")
 
     return detail
+
+
+def filelock(func, env_var: str):
```
agrosu1978, line 57: We should move the initialization of the database structure in a script outside the rs-server. Thus, the python code should not create any table or whatever. It will be easier to maintain the db. I don't see the usage of a filelock when working with a postgres db. If it's a must, than something is wrong somewhere and more problems in future may appear

services/staging/rs_server_staging/main.py
```diff
@@ -55,10 +58,55 @@
...
-def init_db():
+
+def get_config_contents() -> dict:
+    """Return the pygeoapi configuration yaml file contents."""
+    # Open the configuration file
+    with open(get_config_path(), encoding="utf8") as opened:
+        contents = opened.read()
+
+        # Replace env vars by their value
+        contents = os.path.expandvars(contents)
```
agrosu1978: It is about the personal choice here. I would rather use the previous code in which  the loading of the config file is explicitly stopped at the first not found env var, and an appropriate message is printed as well as stopping the execution of the app. The expandvars will just let any unfound env var with that value, and Postgresql Manager will complain later on about this with a traceback exception. As I said, is about the vision, in my case I would go with the def get_manager_def(): function that replaces the placeholders

services/staging/rs_server_staging/main.py
```diff
@@ -72,95 +120,44 @@ def init_db():
...
-        raise RuntimeError(f"Error when trying to read the environment variable: {e}") from e
-    return manager_def
-
+        message = "Error reading the manager definition for pygeoapi PostgreSQL Manager"
+        logger.error(message)
+        raise RuntimeError(message)
+    connection = manager_def["connection"]
+
+    # Create SQL Alchemy engine
+    engine = get_engine(**connection)
```
agrosu1978, line 139: I am ok with getting the engine from the database.py. But again, moving the logic of the database structure outside the pythin code will easiness the maintaining of it as well as this part of the src which will be deleted

agrosu1978 (changes requested):
I agree with:
- moving the config.yaml file to staging.yaml file
- taking the engine from database.py, although it should be completely remove if we decide to use a script to create db structure from the helm chart

I would not use the os.path.expandvar function but instead the previous get_manager_def() (from the branch feat-rspy342/persist-staging-jobs)
The filelock mechanism should be removed once the database structure is moved to a single script that the postgres helm chart is using

services/staging/rs_server_staging/main.py
jgaucher-cs: Is it OK for you if instead of using `os.path.expandvars` we use `(Template('$HOME/stuff').substitute(os.environ)` ? See: https://stackoverflow.com/a/37800939. It will also raise a KeyError at the first unfound env var. And it handles both the `$POSTGRES_HOST` and `${POSTGRES_HOST}` syntax.

services/common/rs_server_common/utils/utils2.py
agrosu1978: We decided to do this in a next sprint

services/staging/rs_server_staging/main.py
agrosu1978: Ok with this approach.

PR #716, opened 2024-11-26
fix: catalog/collections/xxx/queryables

jgaucher-cs:


services/catalog/rs_server_catalog/user_handler.py
```diff
@@ -252,7 +252,7 @@ def remove_user_from_collection(collection: dict, user: str) -> dict:
     Returns:
         dict: The collection without the user ID in the id section.
     """
-    if user in collection["id"]:
+    if user and (user in collection.get("id", "")):
```
jgaucher-cs, line 255: This fixes https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-429
The /queryables endpoints return an `$id` field that contains the endpoint url, not an `id` field that contains the collection id. 
When running the pytest below, the `"/catalog/collections/toto:S1_L1/queryables"` endpoint returns: 
```json
{
  "$id": "http://testserver/collections/toto_S1_L1/queryables",
  "type": "object",
  "title": "STAC Queryables.",
  "$schema": "http://json-schema.org/draft-07/schema#",
  "properties": {
    "id": {
      "$ref": "https://schemas.stacspec.org/v1.0.0/item-spec/json-schema/item.json#/definitions/core/allOf/2/properties/id",
      "title": "Item ID",
      "description": "Item identifier"
    },
    "datetime": {
      "type": "string",
      "title": "Acquired",
      "format": "date-time",
      "pattern": "(\\+00:00|Z)$",
      "description": "Datetime"
    },
    "geometry": {
      "$ref": "https://geojson.org/schema/Feature.json",
      "title": "Item Geometry",
      "description": "Item Geometry"
    }
  },
  "additionalProperties": true
}
```
@vprivat-ads @Padeanu @agrosu1978 do you know if we should remove the `toto_` prefix from the `"$id": "http://testserver/collections/toto_S1_L1/queryables"` value above ?

vprivat-ads (approved):

services/catalog/rs_server_catalog/user_handler.py
agrosu1978: Normally, the name of the collection is S1_L1. I think the toto_ prefix should be removed. We can keep it for now, but normally the queryables should have the same behavior as "/catalog/collections/{owner_id}:{collection_id}" endpoint

PR #725, opened 2024-11-28
Support the http redirections

agrosu1978:
When requesting a file, the response may include an [HTTP 307 redirect](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307). This PR adds support for handling such scenarios by ensuring the Authorization header with the bearer access token is resent when following the redirect. This behavior is necessary because the requests library does not do this automatically  ([on purpose](https://github.com/psf/requests/issues/2949#issuecomment-167664262))


vprivat-ads (approved):
nice!

PR #745, opened 2024-12-04
RSPY560: Improved datetime

Padeanu:


services/common/rs_server_common/utils/utils.py
```diff
@@ -107,18 +107,21 @@ def validate_str_list(parameter: str, handler: ValidatorFunctionWrapHandler) ->
...
     This function checks whether the input interval has a valid format (start_date/stop_date) and
     whether the start and stop dates are in a valid ISO 8601 format.
 
     Args:
-        interval (str): The time interval to be validated, with the following format:
+        date_time (str): The time value to be validated, with the following formats:
             "2024-01-01T00:00:00Z/2024-01-02T23:59:59Z"
+            "../2024-01-02T23:59:59Z"
+            "2024-01-01T00:00:00Z/.."
+            "2024-01-01T00:00:00Z"
```
jgaucher-cs: @Padeanu about the rest of this docstring: 
```
    Returns:
        Tuple[Union[None, datetime], Union[None, datetime]]:
            A tuple containing:
            - start_date (datetime): The start date of the interval.
            - stop_date (datetime): The stop date of the interval.
        Or [None, None] if the provided interval is empty.

    Note:
        - The input interval should be in the format "start_date/stop_date"
        (e.g., "2022-01-01T00:00:00Z/2022-01-02T00:00:00Z").
        - If there is an error, err_code and err_text provide information about the issue.
```
Is the `Returns` still up-to-date ? I think you now return 3 values in the tuple. And I think the Note is not up-to-date, maybe you can remove it since the input format is already explained in the `Args` above.

Padeanu: Updated

services/common/rs_server_common/utils/utils.py
```diff
@@ -111,46 +111,87 @@ def validate_str_list(parameter: str, handler: ValidatorFunctionWrapHandler) ->
...
+        """Check if the string can be converted to a valid datetime."""
+        try:
+            datetime.fromisoformat(date)
+            return True
+        except ValueError:
+            return False
+
+    fixed_date_dt, start_date_dt, stop_date_dt = to_dt([fixed_date, start_date, stop_date])
+
+    if fixed_date_dt and "." not in fixed_date:
```
jgaucher-cs: @Padeanu we saw here: https://github.com/stac-utils/stac-fastapi-pgstac/issues/178 that if the ms .000Z are missing, we don't know if it's because the user didn't set them, or because he set them to 0 but the bug in stac-fastapi-pgstac removed them. So in this case, for now we shouldn't define an automatic interval as you did here. If the user wants all results for a full second, he should define explicitly the interval himself.

services/common/rs_server_common/utils/utils.py
```diff
@@ -111,46 +111,87 @@ def validate_str_list(parameter: str, handler: ValidatorFunctionWrapHandler) ->
...
+    fixed_date_dt, start_date_dt, stop_date_dt = to_dt([fixed_date, start_date, stop_date])
+
+    # if fixed_date_dt and "." not in fixed_date:
+    #     # If miliseconds are not defined, don't set to .000Z create a timeinterval, to gather all products
+    #     # from that milisecond
+    #     start_date_dt = fixed_date_dt.replace(microsecond=0)  # type: ignore
+    #     stop_date_dt = fixed_date_dt.replace(microsecond=999999)  # type: ignore
+    #     fixed_date_dt = None
+    #     return fixed_date_dt, start_date_dt, stop_date_dt
+    if stop_date_dt and "." not in stop_date:
```
jgaucher-cs: @Padeanu this is the same problem here: if the ms are missing from the stop date, we don't know if it's because the user set them to 0 or didn't set them. So for now we should not use this code neither.

PR #751, opened 2024-12-05
Docker: Perform apt upgrade to get security fixes

vprivat-ads:
The apt install already upgrades a few dependencies.

Explicitly upgrade all packages to get as many security fixes as possible.

This is the current outcome for the builder:

```
#8 2.567 Calculating upgrade...
#8 2.942 The following packages will be upgraded:
#8 2.942   base-files bash bsdutils libblkid1 libc-bin libc6 libexpat1 libgnutls30
#8 2.943   libgssapi-krb5-2 libk5crypto3 libkrb5-3 libkrb5support0 libmount1
#8 2.943   libseccomp2 libsmartcols1 libsqlite3-0 libssl3 libsystemd0 libudev1 libuuid1
#8 2.943   mount openssl tar tzdata usr-is-merged util-linux util-linux-extra
#8 2.976 27 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
#8 2.976 Need to get 14.9 MB of archives.
#8 2.976 After this operation, 28.7 kB of additional disk space will be used.
```

And for the image:
```
```

### DIFF
--- a/services/adgs/.github/Dockerfile
+++ b/services/adgs/.github/Dockerfile
@@ -17,7 +17,7 @@
 FROM ghcr.io/rs-python/python:3.11.7-slim-bookworm AS builder
 
 # Install dependencies
-RUN apt update && apt install -y libpq-dev gcc
+RUN apt update && apt upgrade -y && apt install -y libpq-dev gcc
 
 # TODO: remove this after we use an official git version tag for eodag
 RUN apt install -y git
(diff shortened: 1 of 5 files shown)


jgaucher-cs (changes requested):
I think we should rather apt update && apt upgrade the base images that are in https://github.com/orgs/RS-PYTHON/packages. This would take less processing time than doing it in every new docker build. We should do this in a script so we can update them easily for future updates. I can take care of this.

jgaucher-cs:
But maybe @vprivat-ads we can keep your modifications because, after all, if the base images are up-to-date, then running `apt update && apt upgrade` in the client images (as you do in this PR) will have no effect. If new security updates are available, they will be installed in the client images and this is a good thing. In this case we will also have to update the base images. 

For info in the client images we do this: 
```
# Clean everything
RUN rm -rf /tmp/whl /root/.cache/pip /var/cache/apt/archives /var/lib/apt/lists/*
```
I will **not** do it in the base images so the `apt update && apt upgrade` in the client images will run faster.

PR #763, opened 2024-12-12
Fix 568: bad alternate key

Padeanu:


services/staging/rs_server_staging/processors.py
```diff
@@ -842,6 +842,10 @@ def publish_rspy_feature(self, feature: Feature):
         # Publish feature to catalog
         # how to get user? // Do we need user? should /catalog/collection/collectionId/items works with apik?
         publish_url = f"{self.catalog_url}/catalog/collections/{self.catalog_collection}/items"
+        # Iterate over assets, and remove alternate field, if they already have one defined.
+        for asset in feature.assets:
```
jgaucher-cs: Can you write it like this instead ? 
```python
for asset in feature.assets.values():
    if hasattr(asset, "alternate"):
        del asset.alternate  # type: ignore
```

jgaucher-cs: Can you write it like this instead ? 
```python
for asset in feature.assets.values():
    if hasattr(asset, "alternate"):
        del asset.alternate  # type: ignore
```

PR #773, opened 2024-12-20
Fix rspy562/pod restarted

jgaucher-cs:
This PR does the following:

1. Switch the "stac_api_common.py::search" function to a sync "def" function, called from the async /search endpoint to avoid blocking the other endpoints (like /health). Notes:

    a. The /search endpoint is implemented as async in stac_fastapi and therefore cannot be switched to sync.

    b. eodag is sync, so when called from an async function, we must call it in a distinct thread with "await run_in_threadpool" to avoid blocking, see: https://stackoverflow.com/a/71517830

    c. Before this PR I was calling only the eodag init in a distinct thread. We also need to do it with the eodag search. To be more secure, I preferred to do it in a higher-level function = in "stac_api_common.py::search" that calls all the eodag functions.

2. Improve other performances by using a @lru_cache on config files, and only reloading the eodag config from env vars when needed.
3. Implement python function decorators that can be used either with sync or async functions.

We still have a risk that other async endpoints (now or in the future) that call long sync functions take too much time and block other endpoints like /health and crash the pods. After discussing with @agrosu1978, we saw that increasing the number of fastapi workers seems to always resolve this problem. Vincent Gaudissart advises me to run fastapi with gunicorn for uvicorn, that seems to have a better worker management (to be studied), see: https://www.uvicorn.org/deployment/. 

Normally with this PR the /search endpoint should not be blocking anymore, but we should study if other endpoints may be blocking, and if this is the case, use this solution (=more workers). But: 

1. We would still need to run benchmarkings to see if this solution is robust if hundreds of simultaneous requests are made (is this a real use-case ?)

1. Several workers is not compatible with the dev --reload mode. We use the --reload mode on our dev docker images to allow hot source code modification from the pods. Only the tagged production docker images (created from a git tag) don't use --reload and thus could use several workers.

    This means that we sill may have a blocked /health and crashing pods with the dev images. @vprivat-ads are they used at Airbus or do you only use the tagged production images ?

So what I think we should do is this, to be discussed with @vprivat-ads @nleconte-csgroup @agrosu1978 :

1. Check that with this PR and the current fastapi configuration (=a single worker), the /search endpoint does not block /health anymore at Airbus.

1. Use gunicorn in the tagged production docker images to use several fastapi workers, to be more secure. The rs-server Dockerfiles should be updated at the same time as https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-459 

1. Identify what other async endpoints could be blocking (how ? maybe by running a lot of parallel http requests while running the demo notebooks or Airbus integration tests). Then either: 

    a. Switch these endpoints to sync, or 

    b. Run benchmarkings with a lot of requests on the production docker images with several workers to verify that this solution is robust (but the dev images would not be secured)


jgaucher-cs:
This PR does the following:

1. Switch the "stac_api_common.py::search" function to a sync "def" function, called from the async /search endpoint to avoid blocking the other endpoints (like /health). Notes:

    a. The /search endpoint is implemented as async in stac_fastapi and therefore cannot be switched to sync.

    b. eodag is sync, so when called from an async function, we must call it in a distinct thread with "await run_in_threadpool" to avoid blocking, see: https://stackoverflow.com/a/71517830

    c. Before this PR I was calling only the eodag init in a distinct thread. We also need to do it with the eodag search. To be more secure, I preferred to do it in a higher-level function = in "stac_api_common.py::search" that calls all the eodag functions.

2. Improve other performances by using a @lru_cache on config files, and only reloading the eodag config from env vars when needed.
3. Implement python function decorators that can be used either with sync or async functions.

We still have a risk that other async endpoints (now or in the future) that call long sync functions take too much time and block other endpoints like /health and crash the pods. After discussing with @agrosu1978, we saw that increasing the number of fastapi workers seems to always resolve this problem. Vincent Gaudissart advises me to run fastapi with gunicorn for uvicorn, that seems to have a better worker management (to be studied), see: https://www.uvicorn.org/deployment/. 

Normally with this PR the /search endpoint should not be blocking anymore, but we should study if other endpoints may be blocking, and if this is the case, use this solution. But: 

1. We would still need to run benchmarkings to see if this solution is robust if hundreds of simultaneous requests are made (is this a real use-case ?)

1. Several workers is not compatible with the dev --reload mode. We use the --reload mode on our dev docker images to allow hot source code modification from the pods. Only the tagged production docker images (created from a git tag) don't use --reload and thus could use several workers.

    This means that we sill may have a blocked /health and crashing pods with the dev images. @vprivat-ads are they used at Airbus or do you only use the tagged production images ?

So what I think we should do is this, to be discussed with @vprivat-ads @nleconte-csgroup @agrosu1978 :

1. Check that with this PR and the current fastapi configuration (=a single worker), the /search endpoint does not block /health anymore at Airbus.

1. Use gunicorn in the tagged production docker images to use several fastapi workers, to be more secure. The rs-server Dockerfiles should be updated at the same time as https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-459 

1. Identify what other async endpoints could be blocking (how ? maybe by running a lot of parallel http requests while running the demo notebooks or Airbus integration tests). Then either: 

    a. Switch these endpoints to sync, or 
    a. Run benchmarkings with a lot of requests on the production docker images with several workers to verify that this solution is robust (but the dev images would not be secured)

vprivat-ads:
Hello @jgaucher-cs I propose to merge your work in this sprint and we test the point 1 at Airbus before the delivery.

For the next points 2 & 3, I propose to do it later in a new story (RSPY-578) is this ok for you?

vprivat-ads (approved):

PR #776, opened 2024-12-23
Add the usage for RSPY_DASK_STAGING_CLUSTER_NAME 

agrosu1978:
Add also some pytests for this usage. In local mode there is no need for such env var, because the LocalCluster runs in the same memory space, and there is no way to see other local clusters


services/staging/rs_server_staging/processors.py
```diff
@@ -677,14 +680,33 @@ def dask_cluster_connect(self):
...
+                cluster_id = [
+                    cluster.name
+                    for cluster in gateway.list_clusters()
+                    if (
+                        isinstance(cluster.options, dict)
+                        and "cluster_name" in cluster.options
+                        and cluster.options["cluster_name"] == cluster_name
+                    )
+                ][0]
+                self.cluster = gateway.connect(cluster_id)
```
Padeanu, line 698: what if cluster_id is still None?

Padeanu:
```python
cluster_id = next(
    (
        cluster.name
        for cluster in gateway.list_clusters()
        if isinstance(cluster.options, dict)
        and cluster.options.get("cluster_name") == cluster_name
    ),
    None,
)

if cluster_id is None:
    raise ValueError(f"No cluster found with name: {cluster_name}")
```

Padeanu: It's a bit unclear, since you raise keyerror and indexerror at the end.

Maybe remove the whole try-except block, and raise only if cluster_id is none

agrosu1978: I followed the suggestion

PR #795, opened 2025-01-03
Add opentelemetry requests hook to trace HTTP headers

vprivat-ads:
Optionally trace HTTP headers from requests/responses in OpenTelemetry.

Disabled by default (security), to enable it, define OTEL_PYTHON_REQUESTS_TRACE_HEADERS environment variable.

### DIFF
--- a/services/common/rs_server_common/utils/opentelemetry.py
+++ b/services/common/rs_server_common/utils/opentelemetry.py
@@ -28,6 +28,7 @@ from opentelemetry.instrumentation.aws_lambda import AwsLambdaInstrumentor
 from opentelemetry.instrumentation.fastapi import FastAPIInstrumentor
 from opentelemetry.instrumentation.httpx import HTTPXClientInstrumentor
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor  # type: ignore
+from opentelemetry.instrumentation.requests import RequestsInstrumentor
 from opentelemetry.sdk.resources import Resource
 from opentelemetry.sdk.trace import TracerProvider
 from opentelemetry.sdk.trace.export import BatchSpanProcessor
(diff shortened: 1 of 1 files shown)


jgaucher-cs (commented):
@vprivat-ads do we need the same in rs-client-libraries ? Else I'm OK with this PR.

services/common/rs_server_common/utils/opentelemetry.py
nleconte-csgroup, line 1: Is it only intended for local mode or should it be also applicable for the server mode (K8S) ? If so, we should modify the helm charts and the values in the infrastructure.

nleconte-csgroup (commented):
Is it only intended for local mode or should it be also applicable for the server mode (K8S) ? If so, we should modify the helm charts and the values in the infrastructure.

services/common/rs_server_common/utils/opentelemetry.py
vprivat-ads: It's especially needed for cluster deployment, not so much for local mode. But I'm not sure if updating the helm chart is needed, the idea is to leave it disabled all time and only enable it manually from time to time during specific analyses, like the problem we currently have with cadip and we don't understand the http requests made by eodag towards the cadip stations.

@suberti-ads  what do you think?

vprivat-ads:
> @vprivat-ads do we need the same in rs-client-libraries ? Else I'm OK with this PR.

@jgaucher-cs no, I don't see a need on client side for now. It's very easy to update the code locally if needed, without having resort to OpenTelemetry. This change is useful in cluster mode to avoid having to build a new docker image.

services/common/rs_server_common/utils/opentelemetry.py
suberti-ads: I am agree, As this patch allow to see header sent by platform, i think, we can keep helm chart configuration as is without this variable. We will manually update deployment to activate and deactivate it.

PR #854, opened 2025-01-17
Feat rspy544/invalid catalog download links

agrosu1978:
Fixing the invalid catalog download links


alinbutu:
> Fixing the invalid catalog download links

This fix deals with : https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-544
Host had to be also forwarded from the rs-server-staging to rs-server-catalog, also the link contained twice the asset id instead of the item id.

jgaucher-cs (approved):
If you tested it on the cluster with Alex it's OK for me :)

Padeanu (approved):

PR #879, opened 2025-01-23
Staging: use pygeoapi processor signature

vprivat-ads:
In RSPY-321 I asked to create the staging application as an [OGC API Processes Part 1 Core](https://ogcapi.ogc.org/processes/overview.html) implementation based on pygeoapi.

It means using as much pygeoapi code as possible. The second step is to have our staging processor to use the [signature defined and expected by pygeoapi](https://github.com/geopython/pygeoapi/blob/master/pygeoapi/process/base.py):

```python
    def execute(self, data: dict, outputs: Optional[dict] = None
                ) -> Tuple[str, Any]:
        """
        execute the process

        :param data: Dict with the input data that the process needs in order
                     to execute
        :param outputs: `dict` or `list` to optionally specify the subset of
                        required outputs - defaults to all outputs.
                        The value of any key may be an object and include the
                        property `transmissionMode` - defaults to `value`.
        :returns: tuple of MIME type and process response
                  (string or bytes, or dict)
        """
```

This PR also fixes some documentation and typing issues.

Although the formats of both the request and response of the execute endpoint do not comply to the OGC API specification, there is no change for the caller in this PR (rs-demo integration tests are not modified). This will be fixed in a future PR.

Followup of #860.

### DIFF
--- a/services/common/rs_server_common/s3_storage_handler/s3_storage_handler.py
+++ b/services/common/rs_server_common/s3_storage_handler/s3_storage_handler.py
@@ -112,7 +112,12 @@ class CustomSessionRedirect(requests.Session):
     """
 
     def __init__(self, trusted_domains=list[str] | None):
-        """Initialize the CustomSession instance."""
+        """
+        Initialize the CustomSession instance.
+
+        Args:
+            trusted_domains (list): List of allowed hosts for redirection in case of change of protocol (HTTP <> HTTPS).
+        """
         super().__init__()
         self.trusted_domains: list[str] = trusted_domains or []  # List of allowed hosts for redirection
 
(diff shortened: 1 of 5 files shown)


Padeanu:
Ok, probably i'll write a ticket to pygeoapi to modify the signature of init() and execute() to accept **kwargs, thus allowing us to extend it with custom parameters and still be compliant.

Just as a note, we are still not fully compliant since we use async methods, and we basically shadow that execute() function. I remember that, if we remove asynchronous keyword, and fully implement the BaseProcessor there were some serialization problems coming from Dask.

Padeanu (approved):

vprivat-ads:
The **kwargs parameters have just been addressed here: https://github.com/geopython/pygeoapi/issues/1901

PR #906, opened 2025-01-30
Smart oauth2 token handling for staging

mcolin16:
…a file


services/common/rs_server_common/authentication/authentication_to_external.py
```diff
@@ -186,53 +236,100 @@ def get_station_token(external_auth_config: ExternalAuthenticationConfig) -> str
...
+        logger.info(f"Fetching access token from station url: {external_auth_config.token_url}")
+        try:
+            response = requests.post(
+                external_auth_config.token_url,
+                data=prepare_data(external_auth_config),
+                timeout=5,
+                headers=prepare_headers(external_auth_config),
+            )
+            if response.status_code != HTTP_200_OK:
+                logger.error(
```
jgaucher-cs: Could you use rs_server_common.utils.utils2.log_http_exception to avoid copy/pasting the message ?

services/common/rs_server_common/authentication/authentication_to_external.py
```diff
@@ -46,6 +49,59 @@
 ACCESS_TK_KEY_IN_RESPONSE = "access_token"
 HEADER_CONTENT_TYPE = "application/x-www-form-urlencoded"
 
+# Mandatory attributed that should be present in the token dictionary
```
agrosu1978, line 52: Maybe "attributed" should be replaced by "atributes"? There are two places where this word appears. Suggestion:
\# Mandatory attributes that must be present in the token dictionary:
\# Note: Additional attributes are included beyond those returned by the station response.
\# These extra attributes include the creation date of both the access token and refresh token.

services/common/rs_server_common/authentication/authentication_to_external.py
```diff
@@ -171,13 +227,54 @@ class ExternalAuthenticationConfig:  # pylint: disable=too-many-instance-attribu
     trusted_domains: list[str] | None = None
 
 
-def get_station_token(external_auth_config: ExternalAuthenticationConfig) -> str:
+def validate_token_dict(token_dict: Any, config: ExternalAuthenticationConfig):
+    """
+    Check if the token variable contains the mandatory attributes
+
+    Args:
+        token_dict (Any):
```
agrosu1978, line 235: The rs-documentation will complain when the documentation is built with mkdocs:
 token_dict (Any):  -> add a description

services/common/rs_server_common/authentication/authentication_to_external.py
```diff
@@ -171,13 +227,54 @@ class ExternalAuthenticationConfig:  # pylint: disable=too-many-instance-attribu
...
+                )
+                raise TokenDataNotFound(
+                    f"""Token variable attribute {attr} of the station""" f"""{config.station_id} is None !""",
+                )
+
+
+def get_station_token(external_auth_config: ExternalAuthenticationConfig, token_dict: dict) -> dict:
     """
     Retrieve and validate an authentication token for a specific station and service.
+    Thee are two main use cases:
```
agrosu1978, line 265: Typo: "There"

services/common/rs_server_common/authentication/authentication_to_external.py
```diff
@@ -212,26 +340,66 @@ def get_station_token(external_auth_config: ExternalAuthenticationConfig) -> str
...
+    else:
+        # Check that the token variable contains the mandatory elements
+        validate_token_dict(token_dict, external_auth_config)
+
+        # If the current token expires in less than one minute, create a new request to send
+        # to the authorisation server with the refresh token given in the payload of the request
+        current_date = datetime.datetime.now()
+        diff_in_sec = (current_date - token_dict["access_token_creation_date"]).total_seconds()
+
+        logger.info(f"----------- DIFF VAUT: {diff_in_sec}")
```
agrosu1978, line 361: Maybe a more meaningful message should be written in a logger.info. This is not a logger.debug

services/common/rs_server_common/s3_storage_handler/s3_storage_handler.py
```diff
@@ -916,18 +935,51 @@ def s3_streaming_upload(  # pylint: disable=too-many-locals
...
-            auth=auth,
-        )
-        prepared_request = session.prepare_request(request)
+
         while attempt < max_retries:
             try:
                 self.connect_s3()
                 self.logger.info(f"Starting the streaming of {stream_url} to s3://{bucket}/{key}")
+
+                if not token_lock:
```
agrosu1978, line 948: This is a protection against an input param, it should be put at the beginning of the function

services/staging/rs_server_staging/processors.py
```diff
@@ -633,6 +630,12 @@ def dask_cluster_connect(self) -> Client:  # pylint: disable=too-many-branches,t
             retrieve a list of existing clusters.
         - If no clusters are available, it attempts to create a new cluster scheduler.
 
+        Args:
+            staging_station_id: identifier of the station on which we want to stage data.
+            This variable will ve used to define the dask.distributed.Variable object used
+            to create an access_token shared by all of the Dask workers (there will be as many
+            shared variable as station involved in the staging process)
```
agrosu1978, line 639: Typo. Suggestion: 'as the number of stations involved'

services/common/rs_server_common/s3_storage_handler/s3_storage_handler.py
```diff
@@ -916,18 +935,51 @@ def s3_streaming_upload(  # pylint: disable=too-many-locals
...
+                        "Staging dask.distributed.lock object is None but is mandatory for the staging",
+                    )
+
+                # Use locks to allow only one thread from one process from one worker to
+                # access/refresh the shared token dictionary at a time
+                with token_lock:
+                    try:
+                        token_dict = token_info.get()
+                        # Get/refresh the access token if necessary
+                        token_dict = get_station_token(config, token_dict)
```
agrosu1978, line 959: The dask Variable should be updated in the monitoring thread or a new thread created specially for this from the processing.py in the staging. Here, the token should be provided by the streaming task function that fetches the value from the shared dask Variable. The Lock should be also used within the streaming task function (from processors.py) and not here in the s3_storage_handler. This function should not be changed in fact. The former 'auth' param (from the previous implementation of this function) is in fact the updated token fetched and provided by the streaming task function. Reason for this comment: redundant refreshes across multiple workers / redundant running of the get_station_token function 
See bellow the comment for streaming_task function in processors.py

services/staging/rs_server_staging/processors.py
```diff
@@ -100,13 +87,16 @@ def streaming_task(product_url: str, trusted_domains: list[str], auth: str, buck
...
+        # given worker
+
         s3_handler = S3StorageHandler(
             os.environ["S3_ACCESSKEY"],
             os.environ["S3_SECRETKEY"],
             os.environ["S3_ENDPOINT"],
             os.environ["S3_REGION"],
         )
-        s3_handler.s3_streaming_upload(product_url, trusted_domains, auth, bucket, s3_file)
+        s3_handler.s3_streaming_upload(product_url, config, bucket, s3_file, token_dict, token_lock)
```
agrosu1978, line 100: The token_dict is just fetched here from the shared variable. The token_dict should be refreshed by one thread only. This thread should be started before submitting all the dask tasks, similar to the monitoring tasks thread implemented in manage_dask_tasks_results function. 
So here, the code should be something like:
```
try:
        # Create a thread lock to synchronize access to shared resources between the threads of a
        # given worker
        with token_lock:
             # fetch the current value of the Variable. This Variable is updated by one thread only
             # started in processors.py
             auth = token_dict.get("access_token")
        s3_handler = S3StorageHandler(
            os.environ["S3_ACCESSKEY"],
            os.environ["S3_SECRETKEY"],
            os.environ["S3_ENDPOINT"],
            os.environ["S3_REGION"],
        )
        s3_handler.s3_streaming_upload(product_url, trusted_domains, auth, bucket, s3_file)
```
Please see the comment at line  901

LE: After reading some more documentation about the dask Lock and dask Variable and looking in the source  code from dask Variable, I reached at the following conclusion:
- dask Variable uses a centralized, atomic update mechanism, ensuring that .get() always returns a fully stored value (this is done by the dask scheduler). the usage of client.sync() function leaded me to this idea
- locks are not needed for reading because .get() never returns a partially written state.
- only writes need a lock to prevent multiple updates from conflicting. this should be discussed because normally only the thread from the processors.py should write it

So I guess that here (for dask variable reader) the token_lock should not be used in fact

services/staging/rs_server_staging/processors.py
```diff
@@ -859,23 +893,13 @@ async def process_rspy_features(self, catalog_collection: str) -> tuple[str, dic
...
-                f"Failed to retrieve the token needed to connect to the external station {domain}",
-            )
+        external_auth_config = load_external_auth_config_by_domain(domain)
 
         # connect to the dask cluster
         try:
-            dask_client: Client = self.dask_cluster_connect()
-            self.submit_tasks_to_dask_cluster(token, external_auth_config.trusted_domains, dask_client)
+            dask_client = self.dask_cluster_connect(external_auth_config.station_id)
+            self.submit_tasks_to_dask_cluster(external_auth_config, dask_client)
```
agrosu1978, line 903: Before calling the submit function, a new thread should be started (similar to manage_dask_tasks_results):
```
dask_client = self.dask_cluster_connect(external_auth_config.station_id)
try:
    await asyncio.to_thread(self.manage_token, self.external_auth_config, self.manage_token, self.token_lock, self.token_info)
except Exception as e:  # pylint: disable=broad-exception-caught
    self.log_job_execution(JobStatus.failed, 0, f"Error from token refreshment thread: {e}")
self.submit_tasks_to_dask_cluster(external_auth_config, dask_client)
```
The thread should be something like this (pseudocode)
```
def manage_token(self, config, token_lock, token_info):
   self.logger.info("Starting the token refreshment logic")

   while (self.process_is_running):
    with token_lock:
        try:
            token_dict = token_info.get()
            # Get/refresh the access token if necessary
            token_dict = get_station_token(config, token_dict)
            token_info.set(token_dict)
         except ......
              ...........
    time.sleep(TIME_TO_SLEEP)
```
The self.process_is_running is set to False by the monitoring thread manage_dask_tasks_results when it finishes
The reasons for my request are as follows:
- centralized token refresh logic
- reduces the eventual redundant requests since only this process refreshes the token
- dask workers only need to read the token instead of managing the refresh logic
- no changes in the s3_storage_handler module. the logic of this token refreshment should not be present there
- ensures each task gets the latest token dynamically
- important: one setter / multiple readers pattern. easier to manage and maintain the code. 

We can discuss about this and of course, I can help you

agrosu1978 (changes requested):

services/common/rs_server_common/s3_storage_handler/s3_storage_handler.py
```diff
@@ -916,18 +935,51 @@ def s3_streaming_upload(  # pylint: disable=too-many-locals
...
+        trusted_domains = config.trusted_domains
         session = CustomSessionRedirect(trusted_domains)
         self.logger.debug(f"trusted_domains = {trusted_domains}")
-        request = requests.Request(
-            method="GET",
-            url=stream_url,
-            auth=auth,
-        )
-        prepared_request = session.prepare_request(request)
+
```
agrosu1978, line 942: The retrying should be moved to the calling user level. The s3 botocore mechanism of retrying is enough. I will handle this

PR #914, opened 2025-01-31
fix-rspy448-staging-jobs-wrong-id-handling

alinbutu:
Fix for RSPY-448 ( Staging : status for GET jobs/WRONG_ID/results is 200 and not 404 ). See https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-448


services/staging/rs_server_staging/main.py
```diff
@@ -284,10 +286,20 @@ async def execute_process(req: Request, resource: str, data: ProcessMetadataMode
 @router.get("/jobs/{job_id}")
 async def get_job_status_endpoint(job_id: str = Path(..., title="The ID of the job")):
     """Used to get status of processing job."""
-    job = app.extra["process_manager"].get_job(job_id)
-    if job:
-        return job
-    raise HTTPException(status_code=HTTP_404_NOT_FOUND, detail=f"Job with ID {job_id} not found")
+    try:
+        return app.extra["process_manager"].get_job(job_id)
+    except JobNotFoundError:
```
jgaucher-cs: I think we should not return JSONResponse errors, but instead raise Exceptions.

Can you reraise the original error, like this ?
```py
except JobNotFoundError as error:
    raise HTTPException(status_code=HTTP_404_NOT_FOUND, detail=f"Job with ID {job_id} not found") 
        from error
```

services/staging/rs_server_staging/main.py
```diff
@@ -284,10 +286,20 @@ async def execute_process(req: Request, resource: str, data: ProcessMetadataMode
...
+        # Handle case when job_id is not found
+        return JSONResponse(
+            status_code=HTTP_404_NOT_FOUND,
+            content={"title": "No Such Job", "detail": f"Job with ID {job_id} not found"},
+        )
+    except Exception as e:  # pylint: disable=broad-exception-caught
```
jgaucher-cs: Do you really need these lines 297~ 302 ? You could remove them and just let the exception propagate.

services/staging/rs_server_staging/main.py
```diff
@@ -303,21 +315,42 @@ async def get_jobs_endpoint():
...
+        # Catch any other unexpected exceptions
+        return JSONResponse(
+            status_code=HTTP_500_INTERNAL_SERVER_ERROR,
+            content={"title": "Internal Server Error", "detail": str(e)},
+        )
```
jgaucher-cs: Same as above

services/staging/rs_server_staging/main.py
```diff
@@ -303,21 +315,42 @@ async def get_jobs_endpoint():
...
```
jgaucher-cs: Same as above

alinbutu: taken into account

jgaucher-cs (approved):

vprivat-ads (approved):

PR #919, opened 2025-02-03
RSPY-493: Extend filter attributes

Padeanu:


services/common/config/adgs_queryables.yaml
Padeanu, line 1: Update this values, and checktype in stac_api_common

alinbutu (approved):

PR #979, opened 2025-02-13
fix-rspy542-invalid-service-desc-link

alinbutu:
Fix for https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-542 :
fix service-desc link /catalog/api - HTTP 400 "Invalid endpoint." => shall be valid (return http 200 and expected contents)
fix catalog/queryables HTTP 404 "NotFoundError" - when catalog db has no collections (The pgstac get_queryables function returns the intersection of all queryables over all collections if collection_id is None. If no collection in database the sql stored function get_queryables returns NULL) => response shall be empty (empty list of properties) and the additionalProperties field set to true


services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -1344,6 +1345,23 @@ async def dispatch(
...
 
+            # GET: '/catalog/queryables' when no collections in the catalog
+            if (
+                request.method == "GET"
+                and request.scope["path"] == QUERYABLES
+                and not self.request_ids["collection_ids"]
+                and response_content["code"] == "NotFoundError"
+            ):
+                empty_catalog_queryables_response = {
+                    "$id": f"{request.base_url}queryables",
```
Padeanu, line 1358: is request.base_url always ending with a "/"?

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -1344,6 +1345,23 @@ async def dispatch(
...
             self.clear_catalog_bucket(response_content)
 
+            # GET: '/catalog/queryables' when no collections in the catalog
+            if (
+                request.method == "GET"
+                and request.scope["path"] == QUERYABLES
+                and not self.request_ids["collection_ids"]
+                and response_content["code"] == "NotFoundError"
+            ):
+                empty_catalog_queryables_response = {
```
Padeanu, line 1357: Put a comment here: "Return default queryables on /catalog/queryables when there are no collections in catalog.

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -1344,6 +1345,23 @@ async def dispatch(
...
+                    "$id": f"{request.base_url}queryables",
+                    "type": "object",
+                    "title": "STAC Queryables.",
+                    "$schema": "https://json-schema.org/draft-07/schema#",
+                    "properties": {},
+                    "additionalProperties": True,
```
Padeanu, line 1363: Why is additionalProperties set to true?

services/catalog/rs_server_catalog/user_handler.py
```diff
@@ -83,19 +84,19 @@ def reroute_url(  # type: ignore # pylint: disable=too-many-branches,too-many-st
...
         "/",
+        "/catalog",
         "/catalog/",
         "/catalog/search",
         "/catalog/queryables",
-        "catalog/api",
+        "/catalog/api",
         "/catalog/api.html",
         "/catalog/docs/oauth2-redirect",
-        "/catalog/queryables",
```
Padeanu, line 92: Why was this removed, is not used anywhere in request processing?

services/catalog/tests/conftest.py
```diff
@@ -113,6 +113,20 @@ def client_fixture(start_database):  # pylint: disable=missing-function-docstrin
         yield client
 
 
+@pytest.mark.integration
```
Padeanu, line 116: i guess it's pytest.mark.unit?

services/catalog/tests/conftest.py
```diff
@@ -113,6 +113,20 @@ def client_fixture(start_database):  # pylint: disable=missing-function-docstrin
...
+@pytest.fixture(scope="session", name="client_with_empty_catalog")
+def client_empty_catalog_fixture(start_database):  # pylint: disable=missing-function-docstring, unused-argument
```
Padeanu, line 118: remove type-checkers disabling, since you have a docstr and use all arguments.

services/catalog/tests/test_user_handler.py
```diff
@@ -220,6 +220,43 @@ def test_reroute_oauth2(self, request_ids):
         reroute_url(request, request_ids)
         assert request.scope["path"] == "/docs/oauth2-redirect"
 
+    @pytest.mark.parametrize(
+        "path, expected",
+        [
+            ("/catalog/", "/"),
+            ("/catalog", "/"),
```
Padeanu, line 227: does it work with "catalog/" ?

services/catalog/tests/conftest.py
jgaucher-cs: In fact I'm not sure these @pytest.mark.integration or @pytest.mark.unit are used anywhere (to be verified)

services/catalog/rs_server_catalog/user_catalog.py
alinbutu: is as per Jira 542

services/catalog/rs_server_catalog/user_handler.py
alinbutu: is duplicated

services/catalog/rs_server_catalog/user_catalog.py
alinbutu: OK, agree to put a comment , but will be: "Return empty list of properties and additionalProperties set to true on /catalog/queryables when there are no collections in catalog." as per Vincent answer in RSPY-542

services/catalog/rs_server_catalog/user_catalog.py
alinbutu: request.base_url (inside a route handler) always ends with / (enforced by Starlette) https://github.com/encode/starlette/blob/master/starlette/requests.py

TestClient.base_url may not include /, as TestClient is a testing utility that does not fully mimic a real request.

services/catalog/tests/test_user_handler.py
alinbutu: works with both "/catalog/", "/catalog"

Padeanu (approved):

PR #1010, opened 2025-02-25
RSPY 564: limit the limit

Padeanu:


services/cadip/rs_server_cadip/api/cadip_search.py
```diff
@@ -474,7 +474,7 @@ def process_session_search(  # type: ignore # pylint: disable=too-many-arguments
     sortby: str,
     limit: Annotated[
         Union[int, None],
-        Query(gt=0, le=10000, default=1000, description="Pagination Limit"),
+        Query(gt=0, default=1000, description="Pagination Limit"),
```
vprivat-ads: Can you please change the default to 100?

vprivat-ads (changes requested):

services/cadip/rs_server_cadip/api/cadip_search.py
Padeanu: Yes

PR #1046, opened 2025-03-04
fix-rspy439-add-item-in-non-existing-collection

alinbutu:
Fix for [Catalog] Issue while adding items to a non existing collection: https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-439

The issue applies to both updating / adding items ( PUT / POST ) to a non existing collection. HTTPException with status HTTP_404_NOT_FOUND will be raised when the collection in the request does not exist.

### DIFF
--- a/services/catalog/tests/test_endpoints.py
+++ b/services/catalog/tests/test_endpoints.py
@@ -1612,7 +1612,6 @@ class TestCatalogPublishFeatureWithoutBucketTransferEndpoint:
         """
         ENDPOINT: PUT: /catalog/collections/{user:collection}/items/{featureID}
         """
-
         # Change correct feature collection id to match with minimal collection and post it
         a_correct_feature["collection"] = "fixture_collection"
         # Post the correct feature to catalog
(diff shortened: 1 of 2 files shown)


services/catalog/tests/test_endpoints.py
```diff
@@ -1628,13 +1627,70 @@ def test_update_timestamp_feature_fails_with_unfound_item(
...
+        a_correct_feature,
+    ):
+        """
+        ENDPOINT: PUT: /catalog/collections/{user:collection}/items/{featureID}
+        with collection as non existing collection
+        """
+        # Change correct feature collection id to match with minimal collection and post it
+        a_correct_feature["collection"] = "fixture_collection"
+        # Post the correct feature to catalog
+        feature_post_response = client.post(
```
Padeanu: why do you need this post?

In this test you should send a PUT request to a non-existing collection and check that response is 404

alinbutu: Normally an item from a collection should be created before updated. But given the purpose of the test case (to check PUT request for an item in a collection that not exists), indeed does not affect the test behaviour. => Test updated (post deleted)

Padeanu (approved):

jgaucher-cs (approved):

PR #1049, opened 2025-03-05
RSPY 350: staging download uac

Padeanu:


services/common/rs_server_common/authentication/authentication.py
```diff
@@ -194,6 +197,5 @@ def auth_validation(station_type, access_type, *args, **kwargs):  # pylint: disa
     if requested_role not in auth_roles:
         raise HTTPException(
             status_code=status.HTTP_401_UNAUTHORIZED,
-            detail=f"Authorization does not include the right role to {access_type} "
-            f"from the {full_station!r} station",
+            detail="Invalid authorization role",
```
jgaucher-cs: Could you keep a more relevant error message ? e.g. `"Missing {requested_role} authorization role"`. It will be easier for a user to ask the admin to add this role.

services/staging/rs_server_staging/main.py
```diff
@@ -58,14 +65,101 @@
...
+    """Return true if a user must be authenticated to use this endpoint route path."""
+
+    # Remove the /catalog prefix, if any
+    path = route_path.removeprefix("/catalog")
+
+    no_auth = (path in ["/api", "/api.html", "/health", "/_mgmt/ping"]) or path.startswith("/auth/")
+    return not no_auth
+
+
+class AuthenticationMiddleware(BaseHTTPMiddleware):  # pylint: disable=too-few-public-methods
```
jgaucher-cs: This AuthenticationMiddleware is the same as in rs-server/services/catalog/rs_server_catalog/main.py. Could you move it to rs-server/services/common/rs_server_common/utils/utils.py and change the catalog and staging so they use it ?

services/staging/rs_server_staging/main.py
```diff
@@ -58,14 +65,101 @@
...
+
+            # Login and redirect to the calling endpoint.
+            except LoginAndRedirect:
+                return await oauth2.login(request)
+
+        # Call the next middleware
+        return await call_next(request)
+
+
+class HandleExceptionsMiddleware(BaseHTTPMiddleware):  # pylint: disable=too-few-public-methods
```
jgaucher-cs: In rs-server/services/common/rs_server_common/utils/utils.py we already have the class DontRaiseExceptions that does the same. Could you use it instead ? (and adapt it if necessary)

services/staging/rs_server_staging/main.py
```diff
@@ -58,14 +65,101 @@
...
+    oauth2_router = oauth2.get_router(app)
+
+    # Add it to the FastAPI application
+    app.include_router(
+        oauth2_router,
+        tags=["Authentication"],
+        prefix=AUTH_PREFIX,
+        include_in_schema=True,
+    )
+
```
jgaucher-cs: This code is also the same between the staging and catalog, could you move it to rs-server/services/common/rs_server_common/utils/utils.py ?

services/staging/rs_server_staging/main.py
```diff
@@ -335,9 +429,6 @@ async def dask_auth(local_dask_username: str, local_dask_password: str):
         os.environ["LOCAL_DASK_PASSWORD"] = local_dask_password
 
 
-# Configure OpenTelemetry
-opentelemetry.init_traces(app, "rs.server.staging")
```
jgaucher-cs: Why did you remove the opentelemetry tracing ?

Padeanu: Will be reverted

services/staging/rs_server_staging/main.py
Padeanu: This can't be moved, here oauth2 is applied to fastapi app.

PR #1076, opened 2025-03-13
RSPY-588/589/590 - precise staging implementation

mcolin16:


services/staging/rs_server_staging/main.py
```diff
@@ -367,11 +510,14 @@ async def get_specific_job_result_endpoint(request: Request, job_id: str = Path(
     try:
         # Query the database to find the job by job_id
         job = app.extra["process_manager"].get_job(job_id)
-        auth_validation("read", job["process_id"], request=request, staging_process=True)
```
Padeanu, line 370: why was this removed?

services/staging/rs_server_staging/main.py
```diff
@@ -331,34 +452,56 @@ async def get_job_status_endpoint(request: Request, job_id: str = Path(..., titl
...
     """Deletes a specific job from the database."""
     try:
         job = app.extra["process_manager"].get_job(job_id)
-        auth_validation("dismiss", job["process_id"], request=request, staging_process=True)
+    # Handle case when job_id is not found
+    except Exception as error:
+        raise HTTPException(status_code=HTTP_404_NOT_FOUND, detail=f"Job with ID {job_id} not found") from error
+    auth_validation("dismiss", job["processID"], request=request, staging_process=True)
+    try:
+        await validate_request(request)
```
Padeanu: Maybe it's cleaner to  use a validator function for request, and add it to dependecies of each endpoint that needs it.

```python
async def validate_request_dependency(request: Request):
    try:
        await validate_request(request)
    except Exception as e:
        raise HTTPException(status_code=500, detail="Request validation failed") from e
```

and then just

```python
@router.delete(
    "/jobs/{job_id}", 
    dependencies=[Depends(just_for_the_lock_icon), Depends(validate_request_dependency)]
)
```

services/staging/rs_server_staging/main.py
```diff
@@ -367,11 +510,14 @@ async def get_specific_job_result_endpoint(request: Request, job_id: str = Path(
...
-        return JSONResponse(status_code=HTTP_200_OK, content=job["status"])
     except JobNotFoundError as error:
-        # Handle case when job_id is not found
         raise HTTPException(status_code=HTTP_404_NOT_FOUND, detail=f"Job with ID {job_id} not found") from error
+    auth_validation("read", job["processID"], request=request, staging_process=True)
+    try:
+        await validate_request(request)
+        return validate_response(request, job["status"])
+    except Exception as e:
+        raise HTTPException(status_code=HTTP_500_INTERNAL_SERVER_ERROR, detail=str(e)) from e
```
Padeanu: Maybe raise another status code, since it's not an internal server error if user provides a non compliant request.

services/staging/rs_server_staging/staging_endpoints_validation.py
```diff
@@ -0,0 +1,82 @@
...
+    Returns:
+        (dict) dictionary corresponding to the valid staging body
+
+    """
+    if not os.path.isfile(PATH_TO_YAML_OPENAPI):
+        raise FileNotFoundError(f"The following file path was not found: {PATH_TO_YAML_OPENAPI}")
+
+    body = await request.body()
+    openapi_request = StarletteOpenAPIRequest(request, body)
+    OPENAPI.validate_request(openapi_request)
```
Padeanu: What does this OPENAPI.validate_request do? It's not in a try except block, it's not returning any value

services/staging/rs_server_staging/staging_endpoints_validation.py
```diff
@@ -0,0 +1,82 @@
...
+def validate_response(request: Request, data: dict, status_code=HTTP_200_OK) -> Any:
+    """
+    Validate an endpoint response according to the ogc specifications
+    (described as yaml schemas)
+
+    Args:
+        request (Request): input request
+        data (dict): data to send in the endpoint response
+    Returns:
+        (str): return the content of the response as a json string
```
Padeanu: Why return content, and not just  if response is valid (bool)

services/staging/tests/test_staging.py
```diff
@@ -507,6 +671,90 @@ async def test_specific_process(
...
+                },
+            },
+        ),
+        (
+            {
+                "inputs": {
+                    "collection": "Target collection",
+                    "items": {
+                        "href": (
+                            "http://localhost:8002/cadip/search?"
```
Padeanu, line 700: Is this actually tested? In this test you mock the output of whole execute() function.

It would be nice to have a test with a staging body that contains an href and mock the rs-server-cadip response to a valid / invalid FeatureCollection.

services/staging/config/staging_templates/README.md
```diff
@@ -0,0 +1,17 @@
+
+# What are the .yml templates in rs-client staging wrappers used for ?
+
+<p>
+In order to validate the  format of the inputs and outputs of the different endpoint wrappers contained in
+the StagingClient class, we used OGC standards: the OGC process documentation (https://ogcapi.ogc.org/processes) precisely describes what should be the formats of these endpoints inputs and outputs thanks to yaml schemas.
+
+All yaml schemas have been downloaded from this link:
```
Padeanu, line 7: So we need to periodically check if standard was updated and download / update this files?

services/staging/rs_server_staging/processors.py
```diff
@@ -409,11 +404,26 @@ async def execute(
             None: This method doesn't raise any exceptions directly but logs errors if the
                 catalog check fails.
         """
+        # If the content of the staging body is a link STAC itemCollection
+        # (and has no 'value' field containing a STAC ItemCollection)
+        # we launch a request to the corresponding service to load the STAC itemCollection
+        if "items" in data and "href" in data["items"] and "value" not in data["items"]:
```
Padeanu: Maybe a try - expect block here?

https://github.com/RS-PYTHON/rs-server/blob/151227dde5ecb2eeced5cce3ec30177a0fd73a93/services/staging/rs_server_staging/processors.py#L568

services/staging/rs_server_staging/processors.py
```diff
@@ -409,11 +404,26 @@ async def execute(
...
+                headers={"cookie": self.headers.get("cookie", None)},
+                timeout=5,
+            )
+            response.raise_for_status()
+            data["items"]["value"] = response.json()
+
         # self.logger.debug(f"Executing staging processor for {data}")
         item_collection: FeatureCollectionModel | None = (
-            FeatureCollectionModel.parse_obj(data["items"]) if "items" in data else None
+            FeatureCollectionModel.parse_obj(data["items"]["value"])
```
Padeanu, line 452: Maybe here should be similar with this:

https://github.com/RS-PYTHON/rs-server/blob/151227dde5ecb2eeced5cce3ec30177a0fd73a93/services/staging/rs_server_staging/processors.py#L1093

If the search request to rs-server returns an empty item_collection log a message, set job to 100 and stop execution.

services/staging/tests/conftest.py
```diff
@@ -22,6 +22,7 @@
 import asyncio
 import os
 import os.path as osp
+import threading
```
Padeanu, line 25: ?

jgaucher-cs (approved):

poetry.lock
```diff
@@ -1,11 +1,12 @@
-# This file is automatically @generated by Poetry 1.8.5 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.1.1 and should not be changed by hand.
```
vprivat-ads: we should switch to poetry 2 later, in a dedicated story, all of us at the same time

services/staging/config/staging_templates/README.md
vprivat-ads: it's one version every few years. Don't worry

services/staging/config/staging_templates/staging_body.json
```diff
@@ -0,0 +1,52 @@
+{
```
vprivat-ads: we can't retrieve the yaml files from a python dependency rather than copying them in our git repository?

services/staging/pyproject.toml
```diff
@@ -59,6 +59,7 @@ jsonpatch = "^1.33"
 dask-gateway = "2024.1.0"
 # same dask version as in the cluster i.e. same as eopf
 dask = {extras = ["distributed"], version = "2024.5.2"}
+openapi-core = "^0.19.4"
```
vprivat-ads: 0.19.5

services/staging/rs_server_staging/main.py
vprivat-ads: if the error comes from validate_request => Error 400 bad request
If it comes from validate_response => Error 500 internal server error (it's a bug)

services/staging/rs_server_staging/processors.py
```diff
@@ -409,11 +404,26 @@ async def execute(
             None: This method doesn't raise any exceptions directly but logs errors if the
                 catalog check fails.
         """
+        # If the content of the staging body is a link STAC itemCollection
+        # (and has no 'value' field containing a STAC ItemCollection)
+        # we launch a request to the corresponding service to load the STAC itemCollection
+        if "items" in data and "href" in data["items"] and "value" not in data["items"]:
+            response = requests.get(
+                data["items"]["href"],
+                headers={"cookie": self.headers.get("cookie", None)},
```
vprivat-ads: this part shall:
- handle both cookie and x-api-key header
- only if the domain of href is the same as the server's domain (to avoid credentials leakage to the outside world)

vprivat-ads (changes requested):

poetry.lock
jgaucher-cs: @mcolin16 you can run `pip install "poetry<2"` to downgrade to poetry v1 then run `poetry lock` to regenerate this file

services/staging/tests/conftest.py
mcolin16: This is only used in order to create a mock that imitates the threading.Lock object

services/staging/rs_server_staging/main.py
mcolin16: Done

services/staging/rs_server_staging/staging_endpoints_validation.py
mcolin16: It's the openapi_core function that handles the ogc validation of the staging body, it is  supposed to return an exception if the validation fails.I changed the code to put it in a try-except block

agrosu1978 (approved):

poetry.lock
mcolin16: Done

services/staging/config/staging_templates/staging_body.json
mcolin16: The ogc schemas come from the ogcapi-processes repo (https://github.com/opengeospatial/ogcapi-processes/tree/1.0-draft.6.metanorma/core/openapi/schemas)but this repo doesn't have a pip-installable library

PR #1091, opened 2025-03-17
RSPY343: Staging uac

Padeanu:
Todos:

Merge with the other auth tests branch and get staging body for POST request.
Update pyteam / other keys with new roles.
Check swagger
Probably update rs-client to handle UAC 401
Demo


services/staging/rs_server_staging/processors.py
```diff
@@ -873,7 +873,12 @@ async def process_rspy_features(self, catalog_collection: str) -> tuple[str, dic
                     auth_validation,
                 )
 
-                auth_validation(external_auth_config.station_id, "download", request=self.request, staging_process=True)
+                auth_validation(
+                    external_auth_config.station_id,
+                    "staging_download",
```
Padeanu, line 878: ? {resource}_download, tbd

jgaucher-cs: Yes I think it would be safer to use {resource}_download

PR #1111, opened 2025-03-21
Feat/rspy 343 staging uac julien

jgaucher-cs:


services/staging/rs_server_staging/processors.py
```diff
@@ -682,7 +682,7 @@ def dask_cluster_connect(self) -> Client:  # pylint: disable=too-many-branches,t
                 cluster_name = os.environ["RSPY_DASK_STAGING_CLUSTER_NAME"]
 
                 # In local mode, authenticate to the dask cluster with username/password
-                if LOCAL_MODE:
+                if common_settings.LOCAL_MODE:
```
jgaucher-cs, line 685: I'm using common_settings.LOCAL_MODE instead of LOCAL_MODE because it's easier to mock from the pytests

services/staging/tests/conftest.py
```diff
@@ -95,34 +95,6 @@ def client_(mocker):
...
-    """init fastapi client app with auth component."""
-    # Test the FastAPI application, opens the database session
-    mocker.patch("rs_server_staging.main.init_db", return_value=None)
-    mocker.patch("rs_server_staging.main.PostgreSQLManager", return_value=mocker.Mock())
-    # Turn on cluster mode
-    mocker.patch("rs_server_common.settings.LOCAL_MODE", new=False, autospec=False)
-    mocker.patch("rs_server_common.authentication.oauth2.get_user_info", return_value=mocker.Mock())
-
-    if not any(isinstance(m, SessionMiddleware) for m in app.user_middleware):
-        app.add_middleware(SessionMiddleware, secret_key=None)
```
jgaucher-cs, line 109: The problem of this fixture is that it modifies the `app` variable so it has side effects on other tests that also use this `app` variable. I've modified your test below to mock the roles using another method.

services/staging/tests/test_authentication_roles_staging.py
```diff
@@ -47,33 +53,57 @@ def test_auth_roles(mocker, staging_client_auth):
     - Some assertions for execution processes are commented out and need JSON body updates.
     - Mocking is used to simulate database interactions for job allocation.
     """
+    owner_id = "pyteam"
+    test_apikey = False
+    test_oauth2 = True  # test only with the oauth2 cookie
+    await init_test(
```
jgaucher-cs, line 60: Mock the roles for your test

services/staging/tests/test_authentication_staging.py
```diff
@@ -76,7 +88,7 @@ async def test_error_when_not_authenticated(mocker, staging_client, httpx_mock:
                 response = staging_client.request(method, endpoint, json=json_data, **header)
                 logger.debug(response)
                 assert response.status_code not in (
-                    # HTTP_401_UNAUTHORIZED, # 401 is raised by auth_validation if roles are missing
+                    HTTP_401_UNAUTHORIZED,
```
jgaucher-cs, line 92: We had the HTTP_401_UNAUTHORIZED because of the side effect I was talking about above.

services/staging/tests/test_authentication_staging.py
```diff
@@ -91,6 +103,31 @@ async def test_error_when_not_authenticated(mocker, staging_client, httpx_mock:
             else:
                 assert staging_client.request(method, endpoint).status_code == HTTP_401_UNAUTHORIZED
 
+    # Also test the processor rights
```
jgaucher-cs, line 107: Add a check on the staging processor rights

PR #1149, opened 2025-04-01
Update to the new rs-infra repositories

vprivat-ads:
Use the new git repositories following the split of rs-infrastructure:

https://github.com/RS-PYTHON/rs-infra-core
https://github.com/RS-PYTHON/rs-server-deployment
https://github.com/RS-PYTHON/rs-workflow-env

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,9 +15,6 @@
 # Please see the documentation for all configuration options:
 # https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
 
-# WARNING: the ignored dependencies must be the same than in:
-# https://github.com/RS-PYTHON/rs-infrastructure/blob/develop/.github/common/resources/requirements.txt
-# https://github.com/RS-PYTHON/rs-infrastructure/blob/develop/.github/common/resources/requirements-force.txt
 # See https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/controlling-dependencies-updated#ignoring-specific-dependencies
 
 version: 2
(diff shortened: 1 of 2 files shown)


nleconte-csgroup (commented):
Looks OK for me, but python dev team should also confirm as they are the user of the modified resources.

.github/dependabot.yml
```diff
@@ -16,8 +16,8 @@
 # https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
 
 # WARNING: the ignored dependencies must be the same than in:
-# https://github.com/RS-PYTHON/rs-infrastructure/blob/develop/.github/common/resources/requirements.txt
-# https://github.com/RS-PYTHON/rs-infrastructure/blob/develop/.github/common/resources/requirements-force.txt
+# https://github.com/RS-PYTHON/rs-infra-core/blob/develop/.github/common/resources/requirements-dask-eopf.txt
+# https://github.com/RS-PYTHON/rs-infra-core/blob/develop/.github/common/resources/requirements-dask-staging.txt
```
jgaucher-cs: I don't think this comment is relevant anymore. It is not necessary to ignore numpy, pandas, msgpack and toolz because they are not installed anymore in our pyproject.toml files. We should only ignore dask (and maybe dask-gateway and distributed) that is installed in rs-server-staging. 

dask is not installed from the https://github.com/RS-PYTHON/rs-infra-core/blob/develop/.github/common/resources/requirements-dask-xxx.txt files.

resources/sync_config_files.py
```diff
@@ -14,7 +14,7 @@
 
 """
 Create rs-server configuration files from templates.
-Copy them to rs-demo, rs-helm and rs-infrastructure repositories.
+Copy them to rs-demo, rs-helm and rs-server-deployment repositories.
```
jgaucher-cs, line 17: I think this script should be updated to comply with the new rs-infra-core and rs-server-deployment folders... e.g. we use the paths: 
```py
rs_helm_dir / "charts/rs-server-station-secrets/values.yaml")
rs_infra_dir / "rs-server/rs-server-station-secrets/values.yaml")
```
but they are not valid anymore. Should we plan a Jira task for this ?

PR #1200, opened 2025-04-18
Pass api key from staging to catalog

jgaucher-cs:


services/staging/rs_server_staging/processors.py
```diff
@@ -348,10 +348,15 @@ def __init__(
...
-        self.request = credentials
-        self.headers: Headers = credentials.headers
+        self.request = request
         self.stream_list: list[Feature] = []
         #################
+        # Copy authentication headers from original HTTP request
+        self.auth_headers: dict[str, str] = {}
+        for key in APIKEY_HEADER, "cookie":
+            if value := self.request.headers.get(key):
+                self.auth_headers[key] = value
```
jgaucher-cs, line 358: Use the auth_headers in every request we'll make later

services/staging/rs_server_staging/processors.py
```diff
@@ -1215,7 +1217,7 @@ def publish_rspy_feature(self, catalog_collection: str, feature: Feature):
         try:
             response = requests.post(
                 publish_url,
-                headers={"cookie": self.headers.get("cookie", None), "host": self.headers.get("host", None)},
```
jgaucher-cs, line 1218: Here we were not passing the apikey but we were passing the host. I'm not sure the host is needed. What do you think @agrosu1978 @Padeanu ?

services/staging/rs_server_staging/processors.py
vprivat-ads: every request internal to our platform, right?

image

jgaucher-cs: @agrosu1978 @Padeanu can you tell us if the credentials are sent only to internal services ?

services/staging/rs_server_staging/processors.py
agrosu1978: Maybe the host is also contained by self.auth_headers

services/staging/rs_server_staging/processors.py
agrosu1978: yes, only to:
rs-server-catalog
rs-server-cadip
rs-server-auxip

agrosu1978 (approved):

PR #1236, opened 2025-05-06
Feat/RSPY-468: Configure rs-catalog from a Configmap

ecombelles-csgroup:
PR for [https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-468](https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-468)
This PR adds:

- Handler to retrieve data from expiration_bucket file in rs_common (file "s3_storage_config.py"). This handler will be used in rs-server-catalog and rs-server-staging to retrieve a bucket name and a lifespan for a given item configuration. I also added unit tests for this handler.

- Modifications of timestamps_extension to get the lifespan of an item from the config instead of an environment variable.

- Refactor of timestamps_extension

- New unit tests with dedicated test data to check if various items are added with the correct lifespan, following specifications from the Jira ticket. Please note that the step 1 (create buckets) is not done here because we are only interested in the retention time so we keep it as simple as possible.


services/common/rs_server_common/s3_storage_handler/s3_storage_config.py
```diff
@@ -0,0 +1,135 @@
...
+
+FILEPATH_ENV_VAR = "BUCKET_CONFIG_FILE_PATH"
+DEFAULT_FILEPATH = "/app/conf/expiration_bucket.csv"
+
+
+class S3StorageConfigurationError(Exception):
+    """Exception raised when problems occur when retrieving settings from the S3 storage configuration file."""
+
+
+def get_storage_settings_from_config(
```
vprivat-ads, line 102: it looks like this function performs an I/O operation every time, so I'm concerned with a lot of I/O for each catalog operation, I think the config file should be kept in memory, or some cache should be added? We have this kind of things for eodag config I think.

vprivat-ads (changes requested):

agrosu1978 (approved):

services/common/rs_server_common/s3_storage_handler/s3_storage_config.py
ecombelles-csgroup: Good point, I changed the way the data is loaded to address that: now the configuration will be kept in memory using a singleton, and reloaded ONLY if the file it comes from is updated or a new file is given.

vprivat-ads (approved):

PR #1240, opened 2025-05-12
WIP - Fix: Github links in dependencies break pypi publication

ecombelles-csgroup:


.github/workflows/publish-binaries.yml
```diff
@@ -140,16 +140,6 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0 # so that Dunamai produce the correct version
-
-      # Hack the pyproject.toml. TODO: find a cleaner to do this.
```
jgaucher-cs, line 144: Not necessary anymore because I've updated the .toml files. Before: 
```yaml
[tool.poetry.dependencies]
rs-server-common = {path = "../common", develop = true}
```
Now:
```yaml
[tool.poetry.dependencies]
rs-server-common = "*"

[tool.poetry.group.dev.dependencies]
rs-server-common = {path = "../common", develop = true}
```

.github/actions/poetry-install/action.yml
```diff
@@ -32,7 +32,7 @@ runs:
     - uses: ./.github/actions/install-python
     - uses: snok/install-poetry@v1
       with:
-        version: 2.1.0
+        version: 2.1.3
```
jgaucher-cs, line 35: Doesn't work with poetry 2.1.0

pyproject.toml
```diff
@@ -85,6 +86,7 @@ flask-cors = "^5.0.1"
 sphinx = "^8.2.3"
 sphinx-rtd-theme = "^3.0.1"
 sqlalchemy-utils = "^0.41.2"
+click = "8.1.8"
```
jgaucher-cs, line 89: To fix the safety issue

PR #1242, opened 2025-05-13
Poetry 2

Padeanu:


.github/actions/poetry-install/action.yml
```diff
@@ -32,7 +32,7 @@ runs:
     - uses: ./.github/actions/install-python
     - uses: snok/install-poetry@v1
       with:
-        version: 1.8.5 # note: poetry v2 has breaking changes
+        version: 2.1.0 # note: poetry v2 has breaking changes
```
jgaucher-cs: @Padeanu can you remove the comment here ? `# note: poetry v2 has breaking changes`

PR #1368, opened 2025-06-20
RSPY-569: support S3 authentication and download

ecombelles-csgroup:
Implementation of [RSPY-569](https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-569)
Changes in staging processor to support staging from an external s3 bucket.

Changes in this PR: 
    - Changes in the authentication dataclass to handle external s3 credentials retrieved from secrets
    - Refactor: function "prepare_streaming_tasks" moved to tasks.py file
    - AssetInfo is now a dataclass and embeds credentials for assets to be downloaded from external s3
    - Addition of two functions in tasks.py to retrieve credentials from the informations contained in the STAC description of an asset to be staged from a s3 bucket
    - Modifications in processor_staging to skip token handling when staging items from s3 buckets
    - Addition of a new function in s3_storage_handler to stream items from an external s3 bucket with credentials to the usual one
    - Addition and correction of unit tests


jgaucher-cs (approved):

services/common/rs_server_common/s3_storage_handler/s3_storage_handler.py
```diff
@@ -920,18 +908,104 @@ def s3_streaming_upload(  # pylint: disable=too-many-locals
...
+                "s3",
+                endpoint_url=source_endpoint_url,
+                aws_access_key_id=source_access_key,
+                aws_secret_access_key=source_secret_key,
+                use_ssl=True,
+            )
+            presigned_url = source_s3_client.generate_presigned_url(
+                "get_object",
+                Params=source_params,
+                ExpiresIn=3600,  # URL with 1 hour expiration
```
Padeanu: maybe use a variable like here? https://github.com/RS-PYTHON/rs-server/blob/bebce158c636017e3bfa1f8095dd710f305ee2f5/services/catalog/rs_server_catalog/user_catalog.py#L96

@agrosu1978 ?

services/staging/rs_server_staging/processors/processor_staging.py
```diff
@@ -555,14 +524,19 @@ def publish_processed_features(self, catalog_collection: str, refresh_token: Ref
...
+                self.unpublish_rspy_features(catalog_collection, published_features_ids)
                 refresh_token.unsubscribe(self.logger)
                 self.logger.error(f"The item {feature.id} couldn't be published in the catalog")
                 return False
-            published_featurs_ids.append(feature.id)
+            published_features_ids.append(feature.id)
         return True
 
-    def manage_dask_tasks(self, client: Client, catalog_collection: str, refresh_token: RefreshTokenData):
+    def manage_dask_tasks(
```
Padeanu, line 577: Are all (6)
```python
if refresh_token 
```

needed?

services/common/rs_server_common/s3_storage_handler/s3_storage_handler.py
ecombelles-csgroup: Good idea, I'll use the same env var to keep it simple (but with a default value to 3600)

services/staging/rs_server_staging/processors/processor_staging.py
ecombelles-csgroup: Yes because when staging from a s3 bucket we don't have a token at all, so we want to avoid any operation on refresh tokens so it doesn't crash. I haven't found a better way to do it since refresh_token is used here and there in the code and using broader "if" would create a lot of code duplication, but if you have any suggestion to improve it I'm interested

jgaucher-cs (approved):

PR #1373, opened 2025-06-25
RSPY498: Create collection inside staging

Padeanu:
Inside staging, check if catalog collection exists, create it otherwise.

### DIFF
--- a/services/staging/rs_server_staging/processors/processor_staging.py
+++ b/services/staging/rs_server_staging/processors/processor_staging.py
@@ -359,7 +359,7 @@ class Staging(
         self.db_process_manager.update_job(self.job_id, update_data)
         return self._get_execute_result()
 
-    async def check_if_collection_exists(self, catalog_collection):
+    def check_if_collection_exists(self, catalog_collection):
         """
         Checks if a catalog collection exists in the remote catalog service.
         If the collection does not exist (HTTP 404), attempts to create it.
@@ -408,7 +408,7 @@ class Staging(
         Returns:
             bool: True in case of success, False otherwise
         """
-        if not await self.check_if_collection_exists(catalog_collection):
+        if not self.check_if_collection_exists(catalog_collection):
             # Stop catalog check if staging is unable to create the collection
             return False
         # Set the filter containing the item ids to be inserted
--- a/services/staging/rs_server_staging/processors/processor_staging.py
+++ b/services/staging/rs_server_staging/processors/processor_staging.py
@@ -382,9 +382,9 @@ class Staging(
             if response.status_code == 404:
                 # If status is not found, create collection body and try to post it.
                 create_response = requests.post(
-                    collection_url,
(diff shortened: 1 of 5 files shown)


services/staging/rs_server_staging/processors/processor_staging.py
```diff
@@ -356,6 +360,44 @@ def log_job_execution(
         self.db_process_manager.update_job(self.job_id, update_data)
         return self._get_execute_result()
```
ecombelles-csgroup, line 360: For information there is a "collection_exists" function in rs_server_catalog (file user_catalog.py, line 523), and a "create_collection" function in rs_server_common (file stac_api_common.py, line 1008), that are doing similar things as the one you create here, so it may be worth checking if we can refactor all of this to just have 2 functions in rs_server_common for this. It's more a suggestion for later than a change for this PR though

services/staging/rs_server_staging/processors/tasks.py
```diff
@@ -115,3 +115,21 @@ def streaming_task(  # pylint: disable=R0913, R0917
             ) from e
     logger_dask.info(f"The streaming task finished. Returning name of the streamed file {s3_file}")
     return s3_file
+
+
+def get_minimal_collection_body(collection_id):
```
ecombelles-csgroup: In my opinion we should put this kind of function into a "tools.py" file or something similar in utils, in order to keep only functions with the algorithm logic in tasks

services/staging/rs_server_staging/processors/processor_staging.py
```diff
@@ -356,6 +360,44 @@ def log_job_execution(
...
 
+    def check_if_collection_exists(self, catalog_collection):
```
ecombelles-csgroup, line 361: I think this can go in a similar tools.py file like the other function

ecombelles-csgroup (approved):

services/staging/rs_server_staging/processors/processor_staging.py
Padeanu: The create_collection in rs_server_common is reffering to stac-search collections.
The collection_exists from rs_server_catalog is dirrectly calling the pgstac backend so it will work only from catalog pod. We need a solution that is calling the catalog using request.post and forwarding the auth credentials.

services/staging/rs_server_staging/processors/processor_staging.py
Padeanu: Disagree—this method relies on attributes that would need to be passed as parameters if moved to another file.

self.catalog_url, self.auth_headers, self.log_job_execution()

services/staging/rs_server_staging/processors/tasks.py
Padeanu: Ok, will be moved into utils/tools.py

jgaucher-cs (approved):

PR #1374, opened 2025-06-30
Fix SARIF upload category

bogdanap:
Hello from GitHub Code Scanning! 😄👋

We noticed your repo had an unusually large number of active categories for the default branch, which I think was unintended.

Including the docker image version name in the category will create a distinct category for each scan
This means you won't be able to track alert changes between two different runs of the same tool.
Similarly, if an alert is fixed in recent runs, it will continue to appear as open because it's still present in all the other scan categories.

It also causes an exponential increase in the number of open alerts, which makes Code Scanning practically unusable. 

--- 

This PR changes changes the category format from e.g. `ghcr.io/rs-python/rs-server-frontend:0.2a14.post47.dev0` to `rs-python/rs-server-frontend`.

Unfortunately, you'll still be stuck with all the old categories unless you manually delete them via the API.
If it makes things easier, we can clean them up for you. We can either wipe out all the history so you have a clean slate, or target on the misconfigured Trivy categories.

### DIFF
--- a/.github/actions/publish-docker/action.yml
+++ b/.github/actions/publish-docker/action.yml
@@ -155,7 +155,7 @@ runs:
       uses: github/codeql-action/upload-sarif@v3
       with:
         sarif_file: trivy-results-docker.sarif
-        category: ${{ env.docker_image }}
+        category: ${{ inputs.image_basename }}${{ inputs.image_suffix }}
 
     - name: Display link to Trivy results
       run: |


vprivat-ads:
Hi @bogdanap !
Wow, thank you for this unexpected but welcome pull request! Indeed we never managed to get Code Scanning working properly because of this issue, but failed to understand what we were doing wrong as it is the first time we play with this feature.
We'd love if you can wipe out all historical data for us so that we can get a clean state and start to use code scanning as expected :)

vprivat-ads:
@bogdanap, we have 4 other repositories affected, would it be possible to wipe out historical data for all of them?

The complete list would be:
- https://github.com/RS-PYTHON/rs-server
- https://github.com/RS-PYTHON/rs-client-libraries
- https://github.com/RS-PYTHON/rs-dpr-service
- https://github.com/RS-PYTHON/rs-infra-core
- https://github.com/RS-PYTHON/operational-services

bogdanap:
@vprivat-ads Sure, we're happy to clean-up all the mentioned repos. 
I'll get back to you once it's done.

bogdanap:
Ok, I've cleaned up all Code Scanning alerts for the mentioned repos. 
Any new scan will re-create all existing alerts and their status should be accurately tracked from now on.

vprivat-ads:
Thank you @bogdanap ! It seems indeed to work fine now: https://github.com/RS-PYTHON/rs-server/security/code-scanning

We are really happy that you helped us to solve this problem! :1st_place_medal:

PR #1379, opened 2025-07-07
RSPY-622/STAC error management

alinbutu:
https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-622


Padeanu (approved):

services/adgs/rs_server_adgs/fastapi/adgs_app.py
```diff
@@ -34,3 +35,5 @@
 # Set properties for the adgs service
 app.state.get_connection = MockPgstacAdgs.get_connection
 app.state.readpool = MockPgstacAdgs.readpool()
+
+register_stac_exception_handlers(app)
```
jgaucher-cs, line 39: @alinbutu I see that register_stac_exception_handlers is called from auxip and cadip, but not from the catalog. Is it because the errors from the catalog are already wrapped by rs-server/services/common/rs_server_common/middlewares.py::HandleExceptionsMiddleware ?

services/common/rs_server_common/utils/error_handlers.py
```diff
@@ -0,0 +1,53 @@
...
+        """Catch Starlette HTTPExceptions, including 404 from unknown routes."""
+        phrase = HTTPStatus(exc.status_code).phrase
+        code = "".join(word.title() for word in phrase.split())
+
+        # Return STAC-compliant error format
+        payload: ErrorResponse = {
+            "code": code,
+            "description": exc.detail if isinstance(exc.detail, str) else str(exc.detail),
+        }
+        return JSONResponse(status_code=exc.status_code, content=payload)
```
jgaucher-cs: The code from these 2 methods is the same, can you implement it in a single distinct method from the same class ?

services/adgs/rs_server_adgs/fastapi/adgs_app.py
alinbutu: yes , because it does not work with register_stac_exception_handlers in catalog - I think that is because exceptions are caught and transformed in these middlewares (UserCatalogMiddleware  /  HandleExceptionsMiddleware ),  FastAPI’s usual exception handler mechanism (which register_stac_exception_handlers relies on) never receives the exception, so its handlers never get triggered

services/common/rs_server_common/utils/error_handlers.py
alinbutu: ok

jgaucher-cs (approved):

PR #1380, opened 2025-07-15
RSPY-156 data lifecycle (cleanup of expired STAC items)

jgaucher-cs:


services/catalog/rs_server_catalog/data_lifecycle.py
```diff
@@ -0,0 +1,231 @@
...
+            task is deactivated.
+            cancel: Cancel the task
+        """
+        self.logger = Logging.default(__name__)
+        self.app: FastAPI = app
+        self.client_search: CoreCrudClient = client_search
+        self.client_bulk = BulkTransactionsClient()
+        self.s3_handler = S3StorageHandler()
+        self.periodic_task: Task | None = None
+        self.period: float = float(os.getenv("RSPY_DATA_LIFECYCLE_PERIOD", -1))
```
jgaucher-cs: Note: the task will be called automatically from the server-side every n seconds. This is set by an env var. The task is not called from prefect.

services/catalog/rs_server_catalog/response_manager.py
```diff
@@ -1,42 +0,0 @@
-# Copyright 2024 CS Group
```
jgaucher-cs, line 1: This module was not used, I've removed it

services/catalog/rs_server_catalog/timestamps_extension.py
```diff
@@ -38,27 +38,23 @@ def set_timestamps_for_creation(item: dict) -> dict:
     return item
 
 
-def set_timestamps_for_insertion(item: dict, expiration: datetime.datetime | None = None) -> dict:
```
jgaucher-cs, line 41: This parameter was never used, I've removed it

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -1206,27 +1206,6 @@ async def manage_delete_request(self, request: Request):
         await self.build_filelist_to_be_deleted(request)
         return True
 
-    async def retrieve_timestamp(self, request: Request) -> tuple[str, str]:
```
jgaucher-cs, line 1210: This function was never used, I've removed it

services/common/rs_server_common/authentication/oauth2.py
```diff
@@ -315,7 +315,7 @@ async def get_user_info(request: Request) -> AuthInfo:
         # )
         raise HTTPException(
             status.HTTP_401_UNAUTHORIZED,
-            "You cannot login from a console, you need an API key or an OAuth2 session cookie.",
+            f"Error connecting to {str(request.url)!r} without a valid API key or OAuth2 session cookie.",
```
jgaucher-cs, line 318: Nothing to do with this PR but I've updated to a more relevant error message there

services/common/rs_server_common/s3_storage_handler/s3_storage_handler.py
```diff
@@ -159,7 +163,13 @@ class S3StorageHandler:
         s3_client (boto3.client): The s3 client to interact with the s3 storage
     """
 
-    def __init__(self, access_key_id, secret_access_key, endpoint_url, region_name):
+    def __init__(
+        self,
+        access_key_id=os.environ["S3_ACCESSKEY"],
```
jgaucher-cs: There are a lot of places in the rs-server-catalog where we call: 
```py
s3_handler = S3StorageHandler(
  os.environ["S3_ACCESSKEY"],
  os.environ["S3_SECRETKEY"],
  os.environ["S3_ENDPOINT"],
  os.environ["S3_REGION"],
)
```
The constructor is always called with these values, so I've put them as default values so we can simply call
```py
s3_handler = S3StorageHandler()
```

services/common/rs_server_common/s3_storage_handler/s3_storage_handler.py
```diff
@@ -227,20 +238,22 @@ def connect_s3(self):
...
         method to create an S3 client instance using the provided credentials and configuration (see __init__).
         """
-        if self.s3_client is None:
-            self.s3_client = self.__get_s3_client(
-                self.access_key_id,
-                self.secret_access_key,
-                self.endpoint_url,
-                self.region_name,
-            )
+        with self.lock:
```
jgaucher-cs: I've added some thread-safety so 2 threads (that can be called from 2 different endpoint calls by 2 different end-users) will not initialize the same `self.s3_client` instance at the same time. Same thing in the `disconnect_s3` method below.

But this is not sufficient and the S3StorageHandler is not thread safe as a whole: what happens if a first thread calls `disconnect_s3` (which calls `self.s3_client.close()`) while another thread (called by another endpoint by another end-user) is using this `self.s3_client` instance ? The S3 transaction is likely to crash.

**TODO**: I think we should discuss how to make S3StorageHandler thread-safe (in this story or a new Jira story)

services/common/rs_server_common/s3_storage_handler/s3_storage_handler.py
```diff
@@ -289,6 +302,64 @@ def delete_file_from_s3(self, bucket, key, max_retries=S3_MAX_RETRIES):
...
+
+                # The boto3 delete_objects function takes max 1000 items to delete.
+                # Split the key list and process the chunks in parallel.
+                MAX_DELETE = 1000
+                async with asyncio.TaskGroup() as task_group:
+                    for i in range(0, len(keys), MAX_DELETE):
+                        task_group.create_task(
+                            # Do the search in a synchronized thread so we don't block the main thread,
+                            # see: https://stackoverflow.com/a/71517830
+                            run_in_threadpool(
```
jgaucher-cs: As we saw a few months ago with the auxip/cadip search: if an `async` fastapi endpoint calls a `sync` function that takes a lot of time (e.g. an eodag search or a boto3 transaction), all the other calls to other endpoints made by other users will be blocked/frozen until the end of the `sync` call. Calls to the `/health` endpoint made by k8s will be blocked also and this will crash the server !

This is why we need to encapsulate the calls to long `sync` functions inside a `run_in_threadpool` function. I've done it here but this is not done in the other S3StorageHandler functions (or in the endpoints calling the other S3StorageHandler functions). 

**TODO:** I think we should pass the other S3StorageHandler functions to `async` and call the boto3 functions in `run_in_threadpool`. To be discussed (this can be done in this story or another one).

services/catalog/rs_server_catalog/data_lifecycle.py
```diff
@@ -0,0 +1,232 @@
...
+        # Update each item locally and update bucket info
+        for item in items:
+            await self._update_local_item(item, now, bucket_info)
+
+        # Order the items by collection_name
+        items_by_collection: dict[str, list[Item]] = defaultdict(list)
+        for item in items:
+            items_by_collection[item["collection"]].append(item)
+
+        # First, delete all files from the buckets in parallel
```
jgaucher-cs: First I'm deleting all the files from all items from the s3 buckets. Then I'm updating all the items in the stac database.

Does this seem ok to you ? What happens if the deletion from s3 buckets succeeds, but the stac database update fails ? The files will still be in stac but not in the buckets.

Should we do the opposite: first remove from stac, then from the buckets ? In this case, if the stac update succeeds but the bucket removal fails, we won't know anymore what files to delete from the buckets.

pcuq-ads: I would prefer to remove from stac at first, then on the buckets.
There are two reasons for that : 
1.  From end-user point of view, everything starts from the catalog. When the end-user get a STAC item that has expired, he should not try to retrieve asset from the bucket. If you delete the asset without updating the STAC item, the end-user may think there is a problem.
2. We will use an other data lifecycle on OVH Object Storage side to clean up automatically the files on the buckets. Of course, we will set-up the cleaning 24 hours after the expiration delay set on the config map.

In case the feature fails after cleaning up the STAC items, 24 hours after the situation come back to nominal.

services/catalog/rs_server_catalog/data_lifecycle.py
```diff
@@ -0,0 +1,232 @@
...
+        self.periodic_task.cancel()
+        try:
+            await self.periodic_task
+        except Exception:
+            self.logger.error(traceback.format_exc())
+
+    async def run(self):
+        """Trigger the periodic task in a distinct thread and exit."""
+        if (self.period >= 0) and (not self.cancel_flag):
+            self.periodic_task = asyncio.create_task(self._periodic_loop())
```
agrosu1978, line 110: I think this periodic loop is created on the main loop of the fastapi. In case there are a lot of collections to be checked and / or files to be deleted, the catalog main loop will be blocked. To test this, just use a sleep(120) in periodic_once function and try to call another endpoint from stac browser for example. 
The create_task should use asyncio.to_thread in order to run it in a background thread, separated from the fastapi main loop:
asyncio.create_task(
        asyncio.to_thread(self._periodic_loop(), self.period),
    )
 By using self.period in  asyncio.to_thread , you can also remove the await asyncio.sleep(self.period) from the _periodic_loop code

jgaucher-cs: `asyncio.to_thread` is called to call sync functions from async functions without blocking the main loop. This is not the case here because `_periodic_loop` is (and must be) async. 

Note that I use `asyncio.to_thread` in the async `s3_storage_handler.delete_files_from_s3` function to call the sync `s3_client.delete_objects` boto3 method.

services/catalog/rs_server_catalog/timestamps_extension.py
```diff
@@ -38,27 +38,23 @@ def set_timestamps_for_creation(item: dict) -> dict:
...
-        item_eopf_type = item["properties"].get("eopf:type", "*")
-        expiration_range = get_expiration_delay_from_config(item_owner, item_collection, item_eopf_type)
-        expiration_date = datetime.datetime.now() + datetime.timedelta(days=expiration_range)
-        item["properties"]["expires"] = expiration_date.strftime(ISO_8601_FORMAT)
+    item_owner = item["properties"].get("owner", "*")
+    item_collection = item.get("collection", "*").removeprefix(f"{item_owner}_")
+    item_eopf_type = item["properties"].get("eopf:type", "*")
+    expiration_range = get_expiration_delay_from_config(item_owner, item_collection, item_eopf_type)
+    expiration_date = datetime.datetime.now() + datetime.timedelta(days=expiration_range)
+    item["properties"].setdefault("expires", expiration_date.strftime(ISO_8601_FORMAT))
```
jgaucher-cs, line 57: Don't overwrite the expires value coming from the user, if it is set.

services/catalog/rs_server_catalog/timestamps_extension.py
```diff
@@ -76,8 +72,8 @@ def set_timestamps_for_update(item: dict, original_published: str, original_expi
         dict: The updated item.
     """
     item = set_updated_timestamp_to_now(item)
-    item["properties"]["expires"] = original_expires
-    item["properties"]["published"] = original_published
+    item["properties"].setdefault("expires", original_expires)
+    item["properties"].setdefault("published", original_published)
```
jgaucher-cs, line 76: Don't overwrite the expires value coming from the user, if it is set.

services/catalog/tests/test_endpoints/test_catalog_publish_feature_with_bucket_transfer_endpoint.py
```diff
@@ -50,313 +51,238 @@ class TestCatalogPublishFeatureWithBucketTransferEndpoint:
     def test_timestamps_extension_item(
         self,
         client,
+        init_buckets,
         a_correct_feature,
         owner,
         collection_id,
     ):  # pylint: disable=too-many-locals, too-many-arguments
         """Test used to verify that the timestamps extension is correctly set up"""
-        # Create moto server and temp / catalog bucket
```
jgaucher-cs, line 58: Factorize this code in a fixture in conftest. See diffs without whitespaces from https://github.com/RS-PYTHON/rs-server/pull/1380/files?diff=unified&w=1

services/catalog/rs_server_catalog/data_lifecycle.py
jgaucher-cs: Pierre I've done the change you suggested

services/catalog/rs_server_catalog/data_lifecycle.py
```diff
@@ -0,0 +1,259 @@
...
+                try:
+                    # Run the task
+                    await self.periodic_once()
+
+                # Log any error
+                except Exception:  # pylint: disable=broad-exception-caught
+                    self.logger.error(traceback.format_exc())
+
+                # If the caller cancelled execution, we exit the infinite loop
+                if self.cancel_flag:
```
ecombelles-csgroup, line 128: Why this "if" if the condition of the loop is already on cancel_flag?

ecombelles-csgroup (approved):

alinbutu (approved):

services/catalog/rs_server_catalog/data_lifecycle.py
```diff
@@ -0,0 +1,261 @@
...
+
+        # Then, delete all files from the buckets in parallel.
+        # NOTE: if ever this fails, a secondary data lifecycle is set on OVH Object Storage side to clean up
+        # automatically the files on the buckets.
+        # This is done 24 hours after the expiration delay set on the config map.
+        async with asyncio.TaskGroup() as task_group:
+            for bucket_name, bucket_keys in bucket_info.items():
+                # Use a new s3 S3StorageHandler instance for every task as it is not thread-safe
+                task_group.create_task(
+                    S3StorageHandler().adelete_files_from_s3(bucket_name, bucket_keys),
```
vprivat-ads, line 225: "adelete" seems strange, typo?

services/catalog/rs_server_catalog/data_lifecycle.py
jgaucher-cs: It's to exit the loop before calling the sleep.

services/catalog/rs_server_catalog/data_lifecycle.py
jgaucher-cs: adelete = async version of the sync delete function

PR #1382, opened 2025-07-15
Sort assets by PublicationDate

Padeanu:
@vprivat-ads 

I'm not able to test on ops cluster, but this change should sort the assets (but PublicationDate, ascending). Should we sort by this field, or by datetime/published etc?


vprivat-ads:
Hi @Padeanu this sort looks nice, as it matches the default sort implemented by the CADIP itself (ICD chapter 3.4):

> By default, the Files query are to be ordered by Publication Date, in an ascending order (e.g. any query for new Files published since a previous query will find the list ordered from oldest to newest).

We can go with this change :+1:

jgaucher-cs (approved):

vprivat-ads (changes requested):
@Padeanu you can rebase your branch onto develop to get the security fix in poetry lock files, I did a specific PR for this.

vprivat-ads (approved):

PR #1404, opened 2025-07-18
RSPY477: Services content type

Padeanu:


tests/test_authentication_to_external.py
```diff
@@ -1043,7 +1043,7 @@ async def test_set_eodag_auth_token_called_once(  # pylint: disable=too-many-loc
     cadip_session_response,
 ):
     """Test that requesting an auxip/cadip auth token with eodag happens only once per station."""
-    content_type = "application/json"
+    content_type = "application/vnd.oai.openapi+json;version=3.0"
```
Padeanu: maybe this change is not needed, i have to check

services/common/rs_server_common/fastapi_app.py
```diff
@@ -210,4 +210,12 @@ async def lifespan(*_):
             allow_credentials=True,
         )
 
+    @app.middleware("http")
+    async def modify_openapi_content_type(request: Request, call_next):
+        response = await call_next(request)
+        if request.url.path == app.openapi_url:
+            # Replace content-type header
+            response.headers["Content-Type"] = "application/vnd.oai.openapi+json"
```
vprivat-ads: shouldn't we include ";version=3.0" at the end? or is it done magically somewhere else?

Padeanu: Ah, you're right, typo

PR #1407, opened 2025-07-24
Fix/RSPY-709: Support refresh token without "refresh_expires_in" field

ecombelles-csgroup:
Changed token_auth to make refresh_expires_in field not mandatory. In the case when refresh_expires_in is not present, the access_token will be refreshed only using the field "expires_in" but the refresh_token will be refreshed at the same time.

Fix for: https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-709

### DIFF
--- a/services/common/rs_server_common/authentication/token_auth.py
+++ b/services/common/rs_server_common/authentication/token_auth.py
@@ -58,6 +58,11 @@ MANDATORY_TOKEN_ATTRS = [
     "expires_in",
     "refresh_token",
     "refresh_token_creation_date",
+]
+
+# Token attributes that are not mandatory but are expected to be present
+# in a regular token.
+EXPECTED_TOKEN_ATTRS = [
     "refresh_expires_in",
 ]
 
@@ -173,6 +178,15 @@ def validate_token_dict(token_dict: Any, config: StationExternalAuthenticationCo
                 None,
                 TokenDataNotFound,
             )
+    for attr in EXPECTED_TOKEN_ATTRS:
+        if attr not in token_dict:
+            logger.warning(
+                f"Attribute {attr} is not defined in the token variable " f"of the station {config.station_id}.",
+            )
+        if not token_dict[attr]:
+            logger.warning(
+                f"Token variable attribute {attr} of the station {config.station_id} is None.",
+            )
     for attr in "access_token", "refresh_token":
         validate_token_format(attr)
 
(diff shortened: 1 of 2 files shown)


services/common/rs_server_common/authentication/token_auth.py
```diff
@@ -58,6 +58,11 @@ def log_http_exception(*args, **kwargs) -> type[HTTPException]:
     "expires_in",
     "refresh_token",
     "refresh_token_creation_date",
+]
+
+# Token attributes that are not mandatory but are expected to be present
+# in a regular token.
+EXPECTED_TOKEN_ATTRS = [
```
jgaucher-cs: Rather call it OPTIONAL_TOKEN_ATTRS

services/common/rs_server_common/authentication/token_auth.py
```diff
@@ -173,6 +178,15 @@ def validate_token_dict(token_dict: Any, config: StationExternalAuthenticationCo
...
             )
+    for attr in EXPECTED_TOKEN_ATTRS:
+        if attr not in token_dict:
+            logger.warning(
+                f"Attribute {attr} is not defined in the token variable " f"of the station {config.station_id}.",
+            )
+        if not token_dict[attr]:
+            logger.warning(
+                f"Token variable attribute {attr} of the station {config.station_id} is None.",
+            )
```
jgaucher-cs: @vprivat-ads do we want to display a warning if the value is missing from the station ?

services/common/rs_server_common/authentication/token_auth.py
```diff
@@ -261,14 +275,25 @@ def get_station_token(external_auth_config: StationExternalAuthenticationConfig,
...
 
-    # If we have no token yet or if both the access and refresh tokens are expired, we get a new token
-    # using the authorisation grant
-    if not token_dict or (
-        token_dict
-        and (current_date - token_dict["access_token_creation_date"]).total_seconds()
-        > token_dict["expires_in"] - nb_secs_before_token_exp
-        and (current_date - token_dict["refresh_token_creation_date"]).total_seconds()
-        > token_dict["refresh_expires_in"] - nb_secs_before_refresh_token_exp
+    # If we have no token yet
```
jgaucher-cs: To be simplified, try to break it down into several bool variables, something like
```py
if not token_dict:
    get_new_token = True

else:
    if token_dict.get("refresh_expires_in"): # check refresh token
        refresh_token_age = (current_date - token_dict["refresh_token_creation_date"]).total_seconds()
        # do the compare ...
    else: # check access token
        access_token_age = (current_date - token_dict["access_token_creation_date"]).total_seconds()
        expires_in = token_dict["expires_in"]
        # do the compare ...
```

jgaucher-cs (approved):

services/common/rs_server_common/authentication/token_auth.py
ecombelles-csgroup: After feedback I removed the warnings and the EXPECTED_TOKEN_ATTRS that became useless

vprivat-ads (approved):
We just tested this branch and were able to stage a session from Maspalomas with it

PR #1412, opened 2025-07-25
Feat rspy621/catalog download products from s3

agrosu1978:


services/catalog/tests/test_authentication_catalog.py
```diff
@@ -489,6 +493,7 @@ async def test_authentication_and_contents(mocker, httpx_mock: HTTPXMock, client
             **COMMON_FIELDS,
         },
     ]
+    print("DEBUG__")
```
jgaucher-cs: Do you want to keep this ?

services/catalog/tests/test_authentication_catalog.py
```diff
@@ -497,6 +502,7 @@ async def test_authentication_and_contents(mocker, httpx_mock: HTTPXMock, client
 
     # Test a wrong apikey
     if test_apikey:
+        print("YES")
```
jgaucher-cs: And this ?

services/catalog/tests/test_authentication_catalog.py
agrosu1978: Thank you for reminding me about it. No, for sure

services/catalog/tests/test_authentication_catalog.py
agrosu1978: same, thanks !

PR #1514, opened 2025-10-02
RSPY-805: Invalid catalog URL displayed via STAC browser

ecombelles-csgroup:
Jira issue: https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-805

Fix of a bug in the STAC browser in the way the collection names and URL were manipulated, that resulted in broken collection name being created.

~~This fix relies on 2 assumptions:
    - That user names DO NOT contain the symbol "\_"
    - That full collection names are automatically saved in the code as "[Owner_name]\_[Collection_name]"
These assumptions are the only way to identify an owner name in a collection name without having access to a list of owners.~~

**UPDATE:** Objects monitored through the URL formatting contain an "owner" property. This property is now used to extract the owner name from the collection name, removing the need of a blind extraction of that name. In cases when there is no owner field (that should not happen), the collection name is kept as it is.


jgaucher-cs (approved):

Padeanu (approved):

agrosu1978:
> Jira issue: https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-805
> 
> Fix of a bug in the STAC browser in the way the collection names and URL were manipulated, that resulted in broken collection name being created.
> 
> This fix relies on 2 assumptions: - That user names DO NOT contain the symbol "_" - That full collection names are automatically saved in the code as "[Owner_name]_[Collection_name]" These assumptions are the only way to identify an owner name in a collection name without having access to a list of owners.
> 
> To have a more robust fix than this one, we would need to add the possibility to retrieve a complete list of all collection owners in the code.

For the first assumption, I am not that sure. We have to check.

ecombelles-csgroup:
> > Jira issue: https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-805
> > Fix of a bug in the STAC browser in the way the collection names and URL were manipulated, that resulted in broken collection name being created.
> > This fix relies on 2 assumptions: - That user names DO NOT contain the symbol "_" - That full collection names are automatically saved in the code as "[Owner_name]_[Collection_name]" These assumptions are the only way to identify an owner name in a collection name without having access to a list of owners.
> > To have a more robust fix than this one, we would need to add the possibility to retrieve a complete list of all collection owners in the code.
> 
> For the first assumption, I am not that sure. We have to check.

Sorry I forgot to update the comment after last updates, but in the end we don't need to have this restriction on the user names.

services/catalog/rs_server_catalog/user_handler.py
```diff
@@ -270,3 +271,68 @@ def filter_collections(collections: list[dict], prefix: str) -> list[dict]:
...
+    Args:
+        object (dict): The collection
+
+    Returns:
+        dict: The collection passed in parameter with adapted links
+    """
+    user = collection_id = feature_id = ""
+
+    # Case when object is an item
+    if "properties" in object_content and "collection" in object_content:
```
agrosu1978, line 311: I think the owner field is not part of "properties" in case of item.

services/catalog/rs_server_catalog/user_handler.py
```diff
@@ -270,3 +271,68 @@ def filter_collections(collections: list[dict], prefix: str) -> list[dict]:
...
+        if "owner" in object_content["properties"]:
+            user = object_content["properties"]["owner"]
+        object_content = remove_user_from_feature(object_content, user)
+        collection_id = object_content["collection"]
+        feature_id = object_content["id"]
+
+    # Case when object is a collection
+    elif "id" in object_content:
```
agrosu1978, line 317: Can you please test with an empty collection? You will see there is no owner field in this case, how are you going to get the user in this case?

agrosu1978 (approved):

PR #1575, opened 2025-11-12
RSPY-855 - Delete all keys from a folder

Padeanu:


services/common/rs_server_common/s3_storage_handler/s3_storage_handler.py
```diff
@@ -305,42 +305,54 @@ def delete_files_from_s3(self, bucket: str, keys: list[str], max_retries: int =
...
-                    for future in concurrent.futures.as_completed(futures):
-                        future.result()
-
-                # If everything went OK, exit the function
-                return
+                buckets_collection: dict[str, list[str]] = {}
+                for key in keys:
+                    bucket = key.split("/")[2]
+                    if bucket not in buckets_collection:
+                        buckets_collection[bucket] = []
```
jgaucher-cs: Could you remove these two lines and use instead a defaultdict ? 
```
>>> from collections import defaultdict
>>> buckets_collection: dict[str, list[str]] = defaultdict(list)
>>> buckets_collection["a"].append(1)
>>> buckets_collection["b"].extend([3, 4])
>>> buckets_collection.items()
dict_items([('a', [1]), ('b', [3, 4])])
```

services/common/rs_server_common/s3_storage_handler/s3_storage_handler.py
```diff
@@ -305,42 +305,54 @@ def delete_files_from_s3(self, bucket: str, keys: list[str], max_retries: int =
...
-                        for i in range(0, len(keys), MAX_DELETE_FILES)
-                    ]
-                    for future in concurrent.futures.as_completed(futures):
-                        future.result()
-
-                # If everything went OK, exit the function
-                return
+                buckets_collection: dict[str, list[str]] = {}
+                for key in keys:
+                    bucket = key.split("/")[2]
```
jgaucher-cs: Could you use `urlparse` instead ? 
```
>>> from urllib.parse import urlparse
>>> key = "s3://bucket-name/sub/folder/key"
>>> parsed = urlparse(key)
>>> parsed
ParseResult(scheme='s3', netloc='bucket-name', path='/sub/folder/key', params='', query='', fragment='')
>>> bucket = parsed.netloc
```

services/common/rs_server_common/s3_storage_handler/s3_storage_handler.py
```diff
@@ -305,42 +305,54 @@ def delete_files_from_s3(self, bucket: str, keys: list[str], max_retries: int =
...
+                    if bucket not in buckets_collection:
+                        buckets_collection[bucket] = []
+                    path = key.strip().lstrip("/")
+                    s3_files = self.list_s3_files_obj(bucket, path[len(f"s3://{bucket}/") :])
```
jgaucher-cs: Could you use `urlparse` instead as above ?
```python
s3_files = self.list_s3_files_obj(parsed.netloc, parsed.path)
```

services/common/rs_server_common/s3_storage_handler/s3_storage_handler.py
```diff
@@ -305,42 +305,54 @@ def delete_files_from_s3(self, bucket: str, keys: list[str], max_retries: int =
...
+                        # If the key is a folder, expand it with all files inside
+                        buckets_collection[bucket].extend(s3_files)
+
+                for bucket, new_keys in buckets_collection.items():
+                    # Convert the key values into a dict
+                    key_dict = [{"Key": key} for key in new_keys]
+
+                    # The boto3 delete_objects function takes max 1000 items to delete.
+                    # Split the key list and process the chunks in parallel.
+                    with concurrent.futures.ThreadPoolExecutor() as executor:
```
jgaucher-cs: This code is used to delete all files in parallel. If you want to implement it for all buckets, then it should also delete all files from all buckets in parallel. This is not the case here: it deletes all files from the first bucket, waits for the end... then deletes all files from second bucket etc.

You have to switch the order of the `for` loop and the ThreadPoolExecutor. Something like: 
```python
# The boto3 delete_objects function takes max 1000 items to delete.
# Split the key list and process the chunks in parallel.
with concurrent.futures.ThreadPoolExecutor() as executor:

    futures = []
    for bucket, new_keys in buckets_collection.items():

        # Convert the key values into a dict
        key_dict = [{"Key": key} for key in new_keys]

        futures.extend([
            executor.submit(
                self.s3_client.delete_objects,
                Bucket=bucket,
                Delete={"Objects": key_dict[i : i + MAX_DELETE_FILES], "Quiet": True},
            )
            for i in range(0, len(keys), MAX_DELETE_FILES)
        ])

    for future in concurrent.futures.as_completed(futures):
        future.result()
```

services/catalog/rs_server_catalog/data_lifecycle.py
```diff
@@ -218,9 +218,10 @@ async def periodic_once(self, genuine_request: Request | None = None):
         # This is done 24 hours after the expiration delay set on the config map.
         async with asyncio.TaskGroup() as task_group:
             for bucket_name, bucket_keys in bucket_info.items():
+                bucket_files = [f"s3://{bucket_name}/{key}" for key in bucket_keys]
                 # Use a new s3 S3StorageHandler instance for every task as it is not thread-safe
                 task_group.create_task(
-                    S3StorageHandler().adelete_files_from_s3(bucket_name, bucket_keys),
+                    S3StorageHandler().adelete_keys_from_s3(bucket_files),
```
jgaucher-cs: As adelete_keys_from_s3 now accepts several buckets, it's not necessary anymore to create a new task for every bucket. Can you use this instead ?
```python
bucket_files = []

for bucket_name, bucket_keys in bucket_info.items():
    bucket_files.extend([f"s3://{bucket_name}/{key}" for key in bucket_keys])

await S3StorageHandler().adelete_keys_from_s3(bucket_files)
```

jgaucher-cs (approved):

ecombelles-csgroup (approved):

PR #1605, opened 2025-11-24
Update to stac-fastapi-pgstac 6.1.1

vprivat-ads:
Update to:
- https://github.com/stac-utils/stac-fastapi-pgstac/releases/tag/6.1.1

which changes back the format of CORS_ORIGINS, CORS_METHODS and CORS_HEADERS environment variable to the old, simplest, format.

Pull requests:
- https://github.com/RS-PYTHON/rs-server/pull/1605
- https://github.com/RS-PYTHON/rs-helm/pull/183
- https://github.com/RS-PYTHON/rs-demo/pull/217

### DIFF
--- a/services/common/rs_server_common/settings.py
+++ b/services/common/rs_server_common/settings.py
@@ -18,7 +18,6 @@ import os
 from os import environ as env
 
 from httpx import AsyncClient
-from stac_fastapi.pgstac.config import Settings
 from starlette.requests import Request
 
 #########################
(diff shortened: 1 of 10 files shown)


services/common/rs_server_common/settings.py
```diff
@@ -47,9 +46,8 @@ def env_bool(var: str, default: bool) -> bool:
 # Cluster mode is the opposite of local mode
 CLUSTER_MODE: bool = not LOCAL_MODE
 
-# STAC browser URL(s), as seen from the user browser.
-# They are parsed by pydantic from the environment variable CORS_ORIGINS
-CORS_ORIGINS: list[str] = Settings().cors_origins
+# STAC browser URL(s), as seen from the user browser, separated by commas e.g. http://url1,http://url2
+CORS_ORIGINS: list[str] = [url.strip() for url in os.environ.get("CORS_ORIGINS", "").split(",") if url]
```
jgaucher-cs, line 50: I think you can keep the change
```py
CORS_ORIGINS: list[str] = Settings().cors_origins
```
it's cleaner.

vprivat-ads: yes but it adds a direct dependence on stac-fastapi-pgstac in rs_server_common, and there is no technical reason for that, as we don't have a database. In the long term I'd like to remove this dependence to keep only stac-fastapi for rs-server auxip/cadip/prip/edrs. pgstac is only useful for the catalog.

PR #1610, opened 2025-11-28
Improve FTPS client

Padeanu:
In this PR I've moved all methods used for FTP/FTPES interaction in rs-server-common.

By doing this, we can reuse code in rs-server-staging and rs-server-edrs.
EDRSConnector is just a wrapper with some adaptations.

Both rs-server-edrs (used for stac search) and rs-server-staging (streaming) can connect to a EDRS station via FTP or FTPES, supporting client key, certificate and CA certificate. The env key to toggle between in configuration is "USE_SSL".


services/edrs/rs_server_edrs/edrs_connector.py
```diff
@@ -9,28 +9,36 @@
...
 from typing import Any
 
 import xmltodict
-from rs_server_common.utils.logging import Logging
+import yaml
+from rs_server_common.ftp_handler.ftp_handler import FTPClient
 
-logger = Logging.default(__name__)
+# Default path to stations configuration YAML
+DEFAULT_EDRS_STATIONS_CONFIG = ADGS_CONFIG = (
```
jgaucher-cs: Why ADGS_CONFIG ?

jgaucher-cs (approved):

services/edrs/rs_server_edrs/edrs_connector.py
Padeanu: will be removed.

PR #1669, opened 2025-12-19
Handle certs directly from env var

Padeanu:


vprivat-ads:
Close as EDRS is gone.

PR #1670, opened 2025-12-19
update sync config

alinbutu:


resources/sync_config_files.py
```diff
@@ -408,14 +464,22 @@ def copy_to_helm_or_infra(
         try:
             copy_to_helm_or_infra_single_doc(params, output_configs[params.output_doc_index])
         except Exception as e:  # pylint: disable=broad-exception-caught
-            logger.error(f"Failed to update single infra doc: {params}", e)
+            logger.error("Failed to update single infra doc: %s", params, exc_info=e)
```
jgaucher-cs, line 467: @alinbutu why did you replace
```py
f"Failed to update single infra doc: {params}"
```
by
```py
"Failed to update single infra doc: %s", params
```
?

jgaucher-cs (approved):

resources/sync_config_files.py
alinbutu: I think is more robust ,  for the old case something like TypeError: not all arguments converted during string formatting could be thrown

PR #1682, opened 2026-01-07
RSPY-79: Develop PATCH Catalog endpoints

ecombelles-csgroup:
See https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-79

Additions:
- pre-processing of PATCH requests for Items and Collections, to check user's authorizations and update the "updated" timestamp
- Collections are now populated with a "created" and an "updated" timestamp by default on creation


jgaucher-cs (approved):

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -625,6 +625,61 @@ async def manage_search_response(self, request: Request, response: StreamingResp
...
+            original_content = await request.json()
+            content = copy.deepcopy(original_content)
+
+            self._check_user_authorization()
+
+            # Update "updated" timestamp (different field if it is an item or a collection)
+            if "/items/" in request.scope["path"]:
+                content = timestamps_extension.set_updated_timestamp_to_now(content, is_item=True)
+            else:
+                content = timestamps_extension.set_updated_timestamp_to_now(content, is_item=False)
```
vprivat-ads: au choix:

one-liner:
```python
content = timestamps_extension.set_updated_timestamp_to_now(content, is_item="/items/" in request.scope["path"])
```

two-liner:
```python
is_item = "/items/" in request.scope["path"]
content = timestamps_extension.set_updated_timestamp_to_now(content, is_item=is_item)
```

services/catalog/rs_server_catalog/user_catalog.py
```diff
@@ -691,6 +733,17 @@ async def manage_put_post_request(  # pylint: disable=too-many-statements,too-ma
                 content["id"] = owner_id_and_collection_id(self.request_ids["owner_id"], content["id"])
                 if not content.get("owner"):
                     content["owner"] = self.request_ids["owner_id"]
+
+                # See if there is already a collection with this ID. If yes, retrieve its "created" value.
+                try:
+                    existing_collection = await self.client.get_collection(content["id"], request)
+                    date_of_creation = existing_collection.get("created", "")
+                except Exception as e:  # pylint: disable=broad-exception-caught
+                    logger.debug("Collection %s doesn't exist and will be created: %s", content["id"], e)
```
vprivat-ads, line 740: strange to log any exception in debug? if the exception is when the collection doesn't exist, is there not a way to test its existence without exception?

vprivat-ads (approved):

services/catalog/rs_server_catalog/user_catalog.py
ecombelles-csgroup: I know, but this was already implemented like that. If you look at line 457, there is a function that does the same thing to check if a collection exists or not. I just reused the try/except to not run twice the same function call by using 'collection_exists' followed by 'get_collection'.

I will see if I can change the behavior of this function but the 'get_collection' used comes from pgstac so maybe there is no other way to do this

ecombelles-csgroup: So I checked and there is no built-in method to check if a collection exists in CoreCrudClient. The other way would be to retrieve the list of all existing collections and see if the one we're looking for is inside, but I think it's not more efficient

PR #1693, opened 2026-01-13
RSPY-893: Error 401 when trying to stage a CADIP session.

jgaucher-cs:


services/staging/rs_server_staging/processors/processor_staging.py
```diff
@@ -243,10 +244,8 @@ async def execute(  # pylint: disable=too-many-return-statements,arguments-diffe
         )
         catalog_collection: str = data["collection"]
         # In localmode use getpass.getuser() to get PC username
-        # In clustermode, extract username from oauth2 cookie.
-        self.staging_user = (
-            getpass.getuser() if common_settings.LOCAL_MODE else (await oauth2.get_user_info(self.request)).user_login
```
jgaucher-cs, line 248: @Padeanu I think you wrote this code. It works only with oauth2 but not the api key. You have to use instead request.state.user_login that has already been calculated by rs-server/services/common/rs_server_common/authentication/authentication.py

Padeanu (approved):

ecombelles-csgroup (approved):

vprivat-ads (approved):

PR #1694, opened 2026-01-14
Refactor: restructuration of Catalog code

ecombelles-csgroup:
PR to refactor and restructure the code of rs_server_catalog.

Main changes:
- Splitted user_catalog.py into three files under the "middleware" package. One file is the definition of the middleware, one is for the pre-processing of the request and the last one for the post-processing of the response
- Created a "data_management" package for all functions related to the manipulation of Catalog's data. Contains newly created s3_manager, stac_manager, and other existing files were moved there
- Minimal adjustments to other already existing files to move some functions to more appropriate places
- Update of code documentation
- Some fixes in tests affected by the functions moved

Overall, this refactor is focused on making the catalog easier to handle and understand, but the core code was not changed at all and so the behavior of the catalog is the same as before.


jgaucher-cs (approved):

vprivat-ads (approved):
really nice work!

PR #1709, opened 2026-01-20
Feat rspy745/async functions

astanga-cs:
Feat rspy745- async functions


services/catalog/rs_server_catalog/data_management/s3_manager.py
```diff
@@ -349,7 +349,15 @@ def update_stac_item_publication(  # pylint: disable=too-many-locals,too-many-br
         content.update({"collection": f"{user}_{collection_id}"})
         return content, s3_files_to_be_deleted
 
-    def delete_s3_files(self, s3_files_to_be_deleted: list[str]) -> bool:
+    def _parse_s3_path(self, s3_path: str) -> tuple[str, str]:
```
jgaucher-cs: Can you add a small docstring to this method to tell what it does ?

services/catalog/rs_server_catalog/data_management/s3_manager.py
```diff
@@ -365,12 +373,14 @@ def delete_s3_files(self, s3_files_to_be_deleted: list[str]) -> bool:
             logger.error("Failed to create the s3 handler when trying to delete the s3 files")
             return False
 
-        try:
-            self.s3_handler.delete_keys_from_s3(s3_files_to_be_deleted)
-        except RuntimeError as rte:
-            logger.exception(
-                f"Failed to delete keys from s3 bucket. Reason: {rte}. However, the process will still continue !",
-            )
+        for s3_path in s3_files_to_be_deleted:
```
jgaucher-cs: @astanga-cs when we want to delete several files, it's better to call 
```py
delete_keys_from_s3(list_for files)
```
than
```py
for file in list_of_files:
    delete_file_from_s3(file)
```
because if you want to delete 1000 files, you make a single call to `delete_keys_from_s3`, which will also make a single call to the s3 client (`self.s3_client.delete_objects`).

It will be much faster than make 1000 calls to `delete_file_from_s3` i.e. 1000 calls to the s3 client (`self.s3_client.delete_object`)

There is already an async version of delete_keys_from_s3: `async def adelete_keys_from_s3`, can you use it ?

@agrosu1978 with @ecombelles-csgroup we think that we should use consistent method names: 

  * delete_file_from_s3 and delete_file**s**_from_s3
  * OR delete_key_from_s3 and delete_key**s**_from_s3
  * BUT NOT delete_file_from_s3 and delete_key**s**_from_s3 as it is now, what do you think ?

agrosu1978: Yes, agree. I tried to  use the "key" everywhere; this is more natural in the s3 obj storage world

services/catalog/rs_server_catalog/data_management/s3_manager.py
```diff
@@ -360,10 +360,10 @@ def delete_s3_files(self, s3_files_to_be_deleted: list[str]) -> bool:
             return False
 
         try:
-            self.s3_handler.delete_keys_from_s3(s3_files_to_be_deleted)
+            await self.s3_handler.adelete_keys_from_s3(s3_files_to_be_deleted)
         except RuntimeError as rte:
             logger.exception(
-                f"Failed to delete keys from s3 bucket. Reason: {rte}. However, the process will still continue !",
+                f"Failed to delete file from s3 bucket. Reason: {rte}. However, the process will still continue !",
```
jgaucher-cs, line 366: @agrosu1978 @astanga-cs is it really necessary to keep this method ? Why not call directly 
```py
await self.s3_handler.adelete_keys_from_s3
```
instead of 
```py
await self.delete_s3_files
```

If you want to keep this method, should we rename it from `delete_s3_files` to `delete_s3_keys` ?

agrosu1978: I would leave it like it is for now, because actually this is the "application" level that's calling the "sub-module" s3_handler that's in charge with the s3 architecture and 'keys' is the proper way to address internally. At the application level these keys may be seen as files

PR #1712, opened 2026-01-21
Feat rspy906/stac search external ids

alinbutu:


services/catalog/rs_server_catalog/middleware/request_manager.py
```diff
@@ -57,6 +59,149 @@
...
 
+def iter_external_id_parts(raw: Any) -> list[str]:
+    """Split externalIds input (string/list) into clean, comma-separated tokens."""
+    parts: list[str] = []
+    values = raw if isinstance(raw, list) else [raw]
+    for value in values:
+        if value is None:
+            continue
+        # Allow callers to pass a single string with comma-separated ids.
+        for part in str(value).split(","):
```
jgaucher-cs: Just a quick opinion, you could remove 
```py
if value is None:
    continue
```
And use instead:
> for part in str(value **or ""**).split(","):

services/catalog/rs_server_catalog/middleware/request_manager.py
```diff
@@ -57,6 +59,149 @@
...
+                        detail="CQL2 text parser is not available.",
+                    )
+                cql2_text_parser = cast(Callable[[str], Any], parser)
+                # pylint can't infer the callable from getattr; runtime is safe after callable() check.
+                return cql2_text_parser(raw_filter).to_json()  # pylint: disable=not-callable
+            return json.loads(raw_filter)
+        except Exception as exc:  # pylint: disable=broad-exception-caught
+            raise HTTPException(
+                status_code=HTTP_400_BAD_REQUEST,
+                detail="Invalid filter format for externalIds search.",
```
jgaucher-cs: Could you add the input values to the error message ? So the final user can understand better what happened: 
```py
detail=f"Invalid filter format for externalIds search: raw_filter={raw_filter!r}, filter_lang={filter_lang!r}",
```

services/common/rs_server_common/utils/utils.py
```diff
@@ -72,6 +72,62 @@ def validate_str_list(parameter: str) -> list | str:
     return parameter
 
 
+def _iter_external_id_parts(value: str | list[str]) -> Iterable[str]:
+    values = value if isinstance(value, list) else [value]
+    for raw in values:
+        if raw is None:
+            continue
+        for part in (p.strip() for p in str(raw).split(",")):
```
jgaucher-cs: Here also you could remove the first 2 lines and use: 
> for part in (p.strip() for p in str(raw **or ""**).split(",")):

services/catalog/rs_server_catalog/middleware/request_manager.py
alinbutu: done

services/catalog/rs_server_catalog/middleware/request_manager.py
alinbutu: done

services/common/rs_server_common/utils/utils.py
alinbutu: done

jgaucher-cs (approved):

PR #1723, opened 2026-01-27
Fix: Move S3Manager initialization

ecombelles-csgroup:
S3Manager is now initialized only before usage instead of every time the catalog is called

### DIFF
--- a/services/catalog/rs_server_catalog/middleware/request_manager.py
+++ b/services/catalog/rs_server_catalog/middleware/request_manager.py
@@ -293,8 +293,7 @@ field is not permitted also."
                 # try to get the item if it is already part from the collection
                 item = await self._get_item_from_collection(request)
 
-                s3_manager = S3Manager()
-                content, self.s3_files_to_be_deleted = s3_manager.update_stac_item_publication(
+                content, self.s3_files_to_be_deleted = S3Manager().update_stac_item_publication(
                     content,
                     request,
                     self.request_ids,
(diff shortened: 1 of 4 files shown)


services/catalog/rs_server_catalog/middleware/request_manager.py
```diff
@@ -62,9 +63,13 @@ class CatalogRequestManager:
     def __init__(self, client: CoreCrudClient, request_ids: dict[Any, Any]):
         self.client = client
         self.request_ids = request_ids
-        self.s3_manager = S3Manager()
         self.s3_files_to_be_deleted: list = []
 
+    @lru_cache
+    def s3_manager(self):
+        """Creates a cached instance of S3Manager for this class."""
```
jgaucher-cs: In fact it's 
> Creates a cached instance of S3Manager for this class **instance** (self)

services/catalog/rs_server_catalog/middleware/response_manager.py
```diff
@@ -77,9 +78,13 @@ def __init__(
     ):
         self.client = client
         self.request_ids = request_ids
-        self.s3_manager = S3Manager()
         self.s3_files_to_be_deleted = s3_files_to_be_deleted or []
 
+    @lru_cache
```
jgaucher-cs, line 83: Same here, 
> Creates a cached instance of S3Manager for this class **instance** (self)

PR #1737, opened 2026-02-04
RSPY849: Handle geometry crossing antimeridian 2

Padeanu:
After some difficulties installing footprint_facility using pip and current version, i've added a static version into rs_server_common, that will be excluded from tests / coverage / code quality.
https://github.com/eu-cdse/eometadatatool/blob/main/eometadatatool/footprint_facility.py

After these changes, PRIP products with geometry that intersects anti-meridian are reworked and split into MultiPolygons in order to normalize stac browser display, as you can see in attached images.

Single Polygon geometry:

image (5)
MultiPolygon reworked geometry:

image (4)


vprivat-ads (approved):
nice!

PR #1749, opened 2026-02-09
RSPY926: rs-server-osam

Padeanu:
Migrated https://github.com/RS-PYTHON/operational-services as a rs-server service.

Some unit test were joined.


tests/test_authentication.py
```diff
@@ -221,21 +220,21 @@ async def test_endpoints_security(  # pylint: disable=too-many-arguments, too-ma
...
 
-    # Dummy endpoint arguments
     endpoint_params = {
         "collection": "apik_valid_auth",
         "datetime": None,
         "name": None,
     }
 
-    # The user, authenticated with oauth2, can also use an apikey created by another user.
-    # In this case, the apikey authentication has higher priority and should be used.
```
jgaucher-cs, line 238: @Padeanu why did you remove all comments ? I think they're useful to understand what's happening.

tests/test_authentication.py
```diff
@@ -510,31 +513,29 @@ async def test_endpoint_roles(  # pylint: disable=too-many-arguments,too-many-lo
...
     """
-    Test that the api key has the right roles for the http endpoints.
+    Test that the api key or oauth2 authentication has the right roles for the http endpoints.
     """
-    # Mock the uac manager url
-    if test_apikey:
+
+    is_storage_endpoint = endpoint.startswith("/storage/")
+    is_storage_app = fastapi_app.state.router_prefix is None
+    endpoint_exists = not (is_storage_endpoint and not is_storage_app)
```
jgaucher-cs: These 3 lines are very difficult to understand. What does it mean that an endpoint doesn't exist ?

tests/test_authentication.py
```diff
@@ -545,66 +546,64 @@ async def mock_response(user_info: dict):
...
-        # skip for landing_pages since no need for stations.
-        if station == UNKNOWN_CADIP_STATION and "landing_page" not in station_role:
-            message = json.loads(response.content)["description"]
-            assert (
-                response.status_code == status.HTTP_401_UNAUTHORIZED
-                and f"Authorization does not include the right role to download from the 'cadip_{station}' station"
-                in message
-            ) or (
-                response.status_code == status.HTTP_404_NOT_FOUND
-                and f"Unknown CADIP collection: {station!r}" in message
```
jgaucher-cs, line 584: Why are you not testing this anymore ?

jgaucher-cs (changes requested):
Please see my comments

tests/test_authentication.py
```diff
@@ -14,7 +14,6 @@
 
 """Unit tests for the authentication."""
 
-import json
```
jgaucher-cs, line 17: @Padeanu instead of merging the test_authentication from rs_server and osam, maybe it would be better to keep 2 different files. Because the code was already hard to understand. But after merging it becomes again more complex.

Please note that we also have a distinct test_authentication_staging.py. Maybe we should also keep a distinct test_authentication_osam.py.

pyproject.toml
```diff
@@ -72,6 +73,8 @@ rs-server-catalog = {path = "services/catalog", develop = true}
 rs-server-common = {path = "services/common", develop = true}
 rs-server-frontend = {path = "services/frontend", develop = true}
 rs-server-staging = {path = "services/staging", develop = true}
+rs-server-osam = {path = "services/osam", develop = true}
+opentelemetry-instrumentation-asyncio = "^0.60b1"
```
jgaucher-cs: Why do you need opentelemetry-instrumentation-asyncio ? I don't think it was needed in rs-server or osam.

Padeanu: Will be removed

jgaucher-cs (approved):

alinbutu (approved):

jgaucher-cs (approved):

PR #1757, opened 2026-02-16
RSPY939/auxip gte lte - intersect

Padeanu:


jgaucher-cs (approved):

services/common/rs_server_common/stac_cql2.py
```diff
@@ -35,8 +35,8 @@
     "t_finishes": "ll > rl and lh = rh",
     "t_finishedby": "ll < rl and lh = rh",
     "t_equals": "ll = rl and lh = rh",
-    "t_disjoint": "not (ll <= rh and lh >= rl)",
-    "t_intersects": "ll <= rh and lh >= rl",
+    "t_disjoint": "not (ll < rh or ll = rh and lh > rl or lh = rl)",
+    "t_intersects": "ll < rh or ll = rh and lh > rl or lh = rl",
```
ecombelles-csgroup: The problem if you do this is that the change will be applied to every component, even to those that support lte and gte. I think you should change the lines 124 and 125 of this same file instead, like it's done with "is_cadip", and add an "is_auxip" to do the same thing (or change "is_cadip" to "is_cadip_or_auxip")

ecombelles-csgroup (changes requested):

services/common/rs_server_common/stac_cql2.py
Padeanu: @vprivat-ads should this be the general/default behavior, or should we apply it only for specific cases?

vprivat-ads: @Padeanu @ecombelles-csgroup @jgaucher-cs  @astanga-cs  @alinbutu  I made a mistake on this topic regarding OData operators, I'm actually carefully testing each one of them and found out that correct operators are ge/le and not gte/lte, see https://pforge-exchange2.astrium.eads.net/confluence/pages/viewpage.action?pageId=505564386 for my test results, I'm completing them today with PRIP. Basically ge/le works everywhere except for one CADIP station (MPS), I'll see with ESA if we can ask them to solve this issue so that we can avoid this custom logic entirely. Can you please put this story on hold until then? I'm very sorry about that :(

vprivat-ads:
Superseded by https://github.com/RS-PYTHON/rs-server/pull/1806

PR #1772, opened 2026-02-20
Small osam improvements

jgaucher-cs:
Small improvements to osam: 

- Keep the configmap contents in cache so we don't read it every time
- Calculate app.extra["users_info"] if /storage/accounts/update has not been called
- Remove useless function get_keycloak_configmap_values
- Parallel calls to keycloak and obs api for better performance


services/common/rs_server_common/utils/utils.py
```diff
@@ -432,6 +432,7 @@ def run_in_threads(
     func: Callable[..., Any],
     args_list: Sequence[tuple],
     max_workers: int | None = None,
+    raise_exceptions: bool = False,
```
jgaucher-cs, line 435: @vprivat-ads I think you wrote this function. I've added a `raise_exceptions` param to raise immediately in case of exception, instead of returning the exception in the output list.

services/osam/rs_server_osam/utils/tools.py
```diff
@@ -128,6 +129,7 @@ def get_s3_bucket_configuration(cls, config_file_path: str) -> list[list]:
         return cls.bucket_configuration_csv
 
 
+@lru_cache
```
jgaucher-cs, line 132: After discussing with Stéphane, I've added this cache to avoid reading again the configmap file everytime we need it. It will be read only once. Stéphane told me that, when he changes this file manually, he always restart the pod anyway.

services/osam/rs_server_osam/main.py
```diff
@@ -132,7 +133,7 @@ async def app_lifespan(fastapi_app: FastAPI):
     # the trigger for running the logic in the background task
     fastapi_app.extra["users_sync_trigger"] = threading.Event()
     # save info for future requests of endpoint /storage/account/{user}/rights
-    fastapi_app.extra["users_info"] = dict[str, Any]
+    fastapi_app.extra["users_info"] = None
```
jgaucher-cs: `fastapi_app.extra["users_info"] = dict[str, Any]` for typing is wrong, it could be `fastapi_app.extra["users_info"]: dict[str, Any] = {}` but this is not valid in python anyway. So I've removed the typing.

services/osam/rs_server_osam/main.py
```diff
@@ -258,6 +259,14 @@ def __get_user_rights(user):
         HTTPException: If the user is not found in the in-memory Keycloak user store (HTTP 404).
     """
 
+    # If the users info have not been calculated yet (by calling '/storage/accounts/update')
+    if app.extra["users_info"] is None:
```
jgaucher-cs, line 263: I'm initializing the users info in the case when `/storage/accounts/update` has still not be called after the server startup. This is useful for our tests if e.g. we want to call `/storage/account/{user}/rights` without calling  `/storage/accounts/update` first.

services/osam/rs_server_osam/tasks.py
```diff
@@ -126,32 +128,7 @@ def wrapper(*args, **kwargs):
     return decorator
 
 
-@traced_function()
```
jgaucher-cs, line 129: This function was never used except in the pytests, I've removed it.

services/osam/rs_server_osam/tasks.py
```diff
@@ -165,14 +142,22 @@ def build_users_data_map():
...
+                "keycloak_attribute": get_keycloak_handler().get_obs_user_from_keycloak_user(kc_user),
+                "keycloak_roles": [
+                    role["name"] for role in get_keycloak_handler().get_keycloak_user_roles(kc_user["id"])
+                ],
+            },
+        )
+
+    kc_users = [(kc_user,) for kc_user in get_keycloak_handler().get_keycloak_users()]
+    users_info = run_in_threads(build_user, kc_users, MAX_THREAD_COUNT, raise_exceptions=True)
+    return dict(users_info)
```
jgaucher-cs, line 160: The calls to keycloak are slow, so I'm running this in parallel in max 5 threads. In my tests on the CS keycloak, the running time has decreased from ~6 to 2 seconds.

services/osam/rs_server_osam/tasks.py
```diff
@@ -187,35 +172,37 @@ def link_rspython_users_and_obs_users():
         The linking/unlinking logic is currently commented out and should be implemented
         based on specific integration rules.
     """
+    logger.info("Checking the link between keycloak users and ovh accounts. Creating ovh accounts if missing")
```
jgaucher-cs, line 175: Here also I'm running parallel threads to call keycloak and the obs api. The code itself has not changed. Please compare without whitespaces to understand better the code differences: https://github.com/RS-PYTHON/rs-server/pull/1772/changes?w=1#diff-86e6a0c5ce208616951f23d6920813728c48a1de8839481aac052e45877b1709

services/osam/rs_server_osam/tasks.py
```diff
@@ -187,35 +172,37 @@ def link_rspython_users_and_obs_users():
...
-        # After we create OBS account and update keycloak attributes, we need to refresh the keycloak_users and
-        # obs_users lists to have the updated data for the deletion step
-        keycloak_users = get_keycloak_handler().get_keycloak_users()
-        obs_users = get_ovh_handler().get_all_users()
-        for obs_user in obs_users:
-            # If the cloud provider user is not linked with a keycloak account, remove it.
-            delete_obs_user_account_if_not_used_by_keycloak_account(obs_user, keycloak_users)
-    except Exception as e:  # pylint: disable=broad-exception-caught
-        logger.exception("Exception occurred while syncing users")
-        raise RuntimeError("Exception occurred while syncing users") from e
```
jgaucher-cs, line 218: I've removed the try/except because the logging of all exceptions is already handled by HandleExceptionsMiddleware. So no need to log the exceptions 2 times.

services/osam/rs_server_osam/tasks.py
```diff
@@ -440,44 +427,45 @@ def update_s3_rights_lists(s3_rights):  # pylint: disable=too-many-locals
     ]
     statements: list[dict[str, Any]] = []
     for key, block in access_rights_list_keys:  # pylint: disable=too-many-nested-blocks
-        if s3_rights.get(key):
```
jgaucher-cs, line 443: The only change is that I'm using
```py
        if not s3_rights.get(key):
            continue
```
at the top of the for loop, this allows to remove one level of indentation, the code is more readable this way (in my opinion). This also removes one level of complexity in sonarqube.

services/osam/tests/test_tasks.py
```diff
@@ -174,10 +173,10 @@ def test_link_rspython_users_and_obs_users_raises_on_exception(
 
         with patch("rs_server_osam.tasks.get_ovh_handler") as mock_ovh:
             mock_ovh.return_value.get_all_users.side_effect = Exception("OVH API failure")
-            with pytest.raises(RuntimeError) as excinfo:
+            with pytest.raises(Exception) as excinfo:
                 link_rspython_users_and_obs_users()
 
-            assert "Exception occurred while syncing users" in str(excinfo.value)
+            assert "OVH API failure" in str(excinfo.value)
```
jgaucher-cs, line 179: This is because I've remove the try/except in link_rspython_users_and_obs_users.

services/osam/tests/test_tasks.py
```diff
@@ -921,55 +920,3 @@ def test_apply_user_access_policy(
 
     result = apply_user_access_policy(user, access_policy)
     assert result == expected_result
-
-
-@patch("rs_server_osam.tasks.load_configmap_data", return_value=None)
```
jgaucher-cs, line 926: This is because I've removed get_keycloak_configmap_values

services/common/rs_server_common/utils/utils.py
vprivat-ads: I don't remember, it was probably a code refactoring/cleanup. OK with the change.

vprivat-ads (approved):

Padeanu (approved):

PR #1802, opened 2026-03-10
Use moto.mock_aws instead of moto.server.ThreadedMotoServer

jgaucher-cs:
Hi, I had an annoying error while writing my pytests for https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-928 and 931. 

I need to mock a responses.get:

https://github.com/RS-PYTHON/rs-server/blob/238bfe01613e90c512305ea12244ad5f4884465d/services/catalog/tests/conftest.py#L388

from inside a fixture. The easiest way to do this is to use pytest-responses: https://pypi.org/project/pytest-responses/ (Automatically activate responses across your py.test-powered test suite).

But this was not compatible with the way we use moto (mocked s3 bucket): when calling `requests.post(moto_endpoint + "/moto-api/reset", timeout=5)` I had an error.

So I changed a little the way we use moto in the catalog pytests: 
  
  * Use `moto.mock_aws` instead of `moto.server.ThreadedMotoServer`. I use it in a contextmanager, so it is also shut down automatically, no need to shut it down by hand.
  * Use of `monkeypatch.setenv(key, value)` instead of `os.environ[key] = value`. Same thing; it allows to clean up the environment automatically after each test.
  
After this I saw that some tests were working only when executed after other tests: certainly because, before my changes, the moto server was not shutdown, or because some env vars were not reset. So I fixed all that in this PR.

### DIFF
--- a/services/catalog/rs_server_catalog/data_management/s3_manager.py
+++ b/services/catalog/rs_server_catalog/data_management/s3_manager.py
@@ -82,7 +82,7 @@ class S3Manager:
             content (dict): Files to delete
             s3_handler (S3StorageHandler): S3 handler to use. If None given, will do nothing
         """
-        if self.is_catalog_local_mode:
+        if self.is_catalog_local_mode or (not hasattr(content, "get")):
             return
         for asset in content.get("assets", {}):
             # Retrieve bucket name from config using what's in content
--- a/services/catalog/tests/conftest.py
+++ b/services/catalog/tests/conftest.py
@@ -15,17 +15,19 @@
 """Common fixture for catalog service."""
 
 import os
-from collections import namedtuple
 from contextlib import contextmanager
 
-import requests
-from moto.server import ThreadedMotoServer
+import pytest_responses  # pylint: disable=unused-import # noqa: F401 # used to avoid adding @responses.activate
+from _pytest.monkeypatch import (
+    MonkeyPatch,  # see: https://github.com/pytest-dev/pytest/issues/1872#issuecomment-375108891
+)
+from moto import mock_aws
 from rs_server_common.s3_storage_handler.s3_storage_handler import S3StorageHandler
 from rs_server_common.utils.pytest.pytest_authentication_utils import (
     init_app_cluster_mode,
 )
 
-from tests.helpers import clear_aws_credentials, export_aws_credentials
+from tests.helpers import export_aws_credentials
 
 # Init the FastAPI application with all the cluster mode features (local mode=0)
 # Do this before any other imports.
@@ -121,8 +123,9 @@ def client_empty_catalog_fixture(start_database):  # pylint: disable=missing-fun
         # Ensure the catalog is empty by deleting all collections
         response = client.get("/catalog/collections")
         if response.status_code == 200:
-            for collection in response.json().get("collections", []):
-                client.delete(f"/catalog/collections/{collection['owner']}:{collection['id']}")
+            with _init_buckets():
+                for collection in response.json().get("collections", []):
+                    client.delete(f"/catalog/collections/{collection['owner']}:{collection['id']}")
         yield client  # Does NOT trigger setup_database!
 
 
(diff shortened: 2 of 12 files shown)


services/catalog/rs_server_catalog/data_management/s3_manager.py
```diff
@@ -82,7 +82,7 @@ def clear_catalog_bucket(self, content: dict) -> None:
             content (dict): Files to delete
             s3_handler (S3StorageHandler): S3 handler to use. If None given, will do nothing
         """
-        if self.is_catalog_local_mode:
+        if self.is_catalog_local_mode or (not hasattr(content, "get")):
```
jgaucher-cs, line 85: In some cases, content is a str = the error message, not a dict.

services/catalog/tests/test_endpoints/test_catalog_delete_endpoints.py
```diff
@@ -92,7 +92,7 @@ def test_delete_a_foreign_collection(self, client):
         # To be changed with 405 not allowed after UAC
         assert delete_response.status_code == fastapi.status.HTTP_404_NOT_FOUND
 
-    def test_delete_non_empty_collection(self, client, a_minimal_collection, a_correct_feature):
+    def test_delete_non_empty_collection(self, init_buckets, client, a_minimal_collection, a_correct_feature):
```
jgaucher-cs, line 95: We need the mocked buckets for this tests. Before my changes, I guess it was still up from a previous test.

services/catalog/tests/test_endpoints/test_catalog_publish_feature_with_bucket_transfer_endpoint.py
```diff
@@ -110,13 +106,15 @@ def test_timestamps_extension_item(
...
+        monkeypatch,
         client,
         init_buckets,
         a_correct_feature,
         a_minimal_collection,
     ):
         """Test used to verify update of an item to the catalog."""
-        s3_handler = init_buckets.s3_handler
+        s3_handler = init_buckets
+        monkeypatch.setenv("RSPY_LOCAL_CATALOG_MODE", "0")  # Enable bucket transfer
```
jgaucher-cs, line 117: Before my changes, it was not very clear how this env var was set. Now by default, its value is 1, and we set it to 0 only when needed in a few tests.

services/catalog/tests/test_endpoints/test_catalog_publish_feature_with_bucket_transfer_endpoint.py
```diff
@@ -506,64 +500,32 @@ def test_generate_download_presigned_url(self, client, init_buckets):
...
-            "may24C355000e4102500n.tif",
-        )
-        assert response.status_code == HTTP_500_INTERNAL_SERVER_ERROR
-        assert response.content == b'{"code":"InternalServerError","description":"Failed to find s3 credentials"}'
-
     @pytest.mark.unit
-    def test_failure_while_moving_files_between_buckets(self, client, mocker, a_correct_feature):
+    def test_failure_while_moving_files_between_buckets(self, init_buckets, client, mocker, a_correct_feature):
         """Test failure in transferring files between buckets."""
-        moto_endpoint = "http://localhost:8088"
```
jgaucher-cs, line 522: The init of the mocked buckets is already done by `init_buckets`. The rest of the function has not changed, it's only the indent.

services/catalog/tests/conftest.py
```diff
@@ -131,40 +131,21 @@ def _init_buckets():
     """Initialize s3 moto server and create buckets"""
 
     # Create moto server and temp / catalog bucket
-    moto_endpoint = "http://localhost:8077"
     export_aws_credentials()
-    secrets = {"s3endpoint": moto_endpoint, "accesskey": None, "secretkey": None, "region": ""}
-    # Enable bucket transfer
-    os.environ["RSPY_LOCAL_CATALOG_MODE"] = "0"
-    server = ThreadedMotoServer(port=8077)
```
jgaucher-cs, line 139: Use the mock_aws context manager instead of ThreadedMotoServer

services/catalog/tests/helpers.py
```diff
@@ -112,18 +115,15 @@ def export_aws_credentials():
     Raises:
         None
     """
-    with open(RESOURCES_FOLDER / "s3" / "s3.yml", encoding="utf-8") as f:
-        s3_config = yaml.safe_load(f)
-        os.environ.update(s3_config["s3"])
-        os.environ.update(s3_config["boto"])
+    monkeypatch = MonkeyPatch()
```
jgaucher-cs, line 118: Use monkeypatch instead of modifying os.environ

services/catalog/tests/test_authentication_catalog.py
```diff
@@ -349,28 +343,6 @@ def upload_object():
     upload_object()
     yield {"object_content": object_content, "upload_object": upload_object}
 
-    # Clear bucket at the end of the scope
```
jgaucher-cs, line 352: We don't need anymore to stop and clear the mocked bucket manually, because it's done automatically by the mock_aws context manager.

services/common/tests/s3_storage_handler/helpers.py
```diff
@@ -44,11 +44,3 @@ def export_aws_credentials():
     with open(RESOURCES_FOLDER / "s3" / "s3.yml", encoding="utf-8") as f:
         s3_config = yaml.safe_load(f)
         os.environ.update(s3_config["s3"])
-
-
-def clear_aws_credentials():
```
jgaucher-cs, line 49: We don't need anymore to clean the os.environ manually, because it's done automatically by monkeypatch.

PR #1803, opened 2026-03-11
Validation for self link in /collections

astanga-cs:
feat-rspy464-invalid-self-implicit-mode


Padeanu (commented):
Add pytest

services/catalog/rs_server_catalog/middleware/response_manager.py
```diff
@@ -402,6 +402,21 @@ async def _manage_get_response_content(  # pylint: disable=too-many-locals, too-
             "/collections" in request.scope["path"] and "/items" not in request.scope["path"]
         ):  # /catalog/collections/owner_id:collection_id
             content = adapt_object_links(content, self.request_ids["owner_id"])
+
+            # Self-links shall match the requested URL, even in implicit mode (owner not specified)
+            if (
+                request.url.path.replace(":", "_") != request.scope["path"]
+                and isinstance(content, dict)
+                and not common_settings.CLUSTER_MODE
```
jgaucher-cs: Could you use `common_settings.LOCAL_MODE` instead of `not common_settings.CLUSTER_MODE` ?

This code should only run in local mode, not cluster mode ?

services/catalog/rs_server_catalog/middleware/response_manager.py
```diff
@@ -402,6 +402,21 @@ async def _manage_get_response_content(  # pylint: disable=too-many-locals, too-
...
+            ):
+                self_link = next((s for s in (content.get("links") or []) if s.get("rel") == "self"), None)
```
jgaucher-cs, line 411: Could you add a comment saying what this line does ? `next((s for s in (content.get("links") or []) if s.get("rel") == "self"), None)` or break it into several lines, because it's very hard to understand what it does.

services/catalog/rs_server_catalog/middleware/response_manager.py
```diff
@@ -402,6 +402,21 @@ async def _manage_get_response_content(  # pylint: disable=too-many-locals, too-
...
+                if self_link:
+                    parsed = urlparse(self_link["href"])
+                    if m := re.match(CATALOG_COLLECTIONS + r"/.+?:(?P<collection_id>.*)", parsed.path):
```
jgaucher-cs, line 415: Here also could you add a comment saying what this line does ?  `re.match(CATALOG_COLLECTIONS + r"/.+?:(?P<collection_id>.*)", parsed.path)`

Maybe with an example to help us understand what it does.

jgaucher-cs (changes requested):

jgaucher-cs (approved):

PR #1835, opened 2026-03-26
RSPY-975: Same algorithm to choose staging dask cluster in local and cluster mode

alinbutu:
https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-975


jgaucher-cs (approved):

services/staging/rs_server_staging/processors/processor_staging.py
```diff
@@ -831,32 +831,25 @@ def dask_cluster_connect(
...
-                            cluster.name
-                            for cluster in clusters
-                            if isinstance(cluster.options, dict) and cluster.options.get("cluster_name") == cluster_name
-                        ),
-                        None,
-                    )
-                    self.logger.info(f"Cluster id vaut: {cluster_id}")
+                self.logger.info(f"Requested cluster name: {cluster_name}")
+
+                for cluster in clusters:
```
ecombelles-csgroup: I don't understand why there is this for loop, is is just for logs? I don't see the "is_equal" being used anywhere else
And if so, can it be removed or replaced with a more simple log?

alinbutu: Ok. yes , that loop was pre-existing and was only used for logs. I removed it and replaced it with a single summary log.

PR #1908, opened 2026-06-16
Fix rspy1081/tcontains query fails

astanga-cs:


services/common/rs_server_common/stac_cql2.py
```diff
@@ -111,8 +142,8 @@ def temporal_op_query(op: str, args: list[dict], temporal_mapping: dict[str, str
     if not temporal_mapping:
         raise ValueError("Undefined temporal property mapping")
 
-    props: list[dict] = args[0]["interval"] if "interval" in args[0].keys() else [args[0]]
-    rrange = parse_dtrange(args[1])
+    props = extract_properties(args)
```
alinbutu, line 170: According to time:intervalContains, the left operand is the interval that contains the right operand: T1 starts before T2 starts, and T1 ends after T2 ends. 
Therefore, for t_contains(interval, published), the generated filter should be equivalent to PublicationDate > interval_start AND PublicationDate < interval_end, 
not the impossible condition PublicationDate < interval_start AND PublicationDate > interval_end.
=>
Make extract_interval return both the literal interval and its argument position. Add a helper to swap left/right temporal operands in an operation template.
Use these in temporal_op_query: 
after extracting props and interval, detect whether the literal interval was passed as the left operand. 
If so, swap the operation template before building outq, because the existing replacement logic assumes props is the left side and rrange is the right side.

services/common/rs_server_common/stac_cql2.py
```diff
@@ -95,6 +95,37 @@ def parse_dtrange(  # noqa: C901 # pylint: disable=too-many-branches
...
+def extract_properties(args):
+    """extract property"""
+    props = [a for a in args if isinstance(a, dict) and "property" in a]
+
+    if not props:
+        for a in args:
+            if isinstance(a, dict) and "interval" in a and isinstance(a["interval"], list):
+                props.extend(p for p in a["interval"] if isinstance(p, dict) and "property" in p)
+
+    return props
```
alinbutu, line 129: extract_properties: raise ValueError("No temporal property in args") when no property is found.

alinbutu (approved):

PR #1923, opened 2026-06-30
RSPY-1106: Update to stac-fastapi(-pgstac) 6.3.x

ecombelles-csgroup:
See: https://pforge-exchange2.astrium.eads.net/jira/browse/RSPY-1106

Update of stac-fastpi and stac-fastapi-pgstac to 6.3.x, with update of the imports that changed and the stac_api_common file


services/common/rs_server_common/stac_api_common.py
```diff
@@ -296,15 +296,16 @@ async def fetchval(self, query, *args, column=0, timeout=None):  # pylint: disab
...
             all_collections = filter_allowed_collections(self.all_collections(), self.service, self.request)
             for collection in all_collections:
                 # Add summaries for each collection that has query
                 # This branch is typically executed first in the normal request flow
                 # Because of that, we do not perform additional checks for existing summaries
                 if (query := collection.get("query")) and (summaries := build_summaries(self.service, query)):
                     collection["summaries"] = summaries
-            return all_collections
+            collections: dict[str, list] = {"collections": all_collections, "links": []}
+            return collections
```
vprivat-ads, line 308: I don't understand the code, does it return all collections? If that's the case this is not correct, the collection search shall honor the search query: If I search sentinel-1 collections I shall not receive sentinel-2 collections.

ecombelles-csgroup: I'm not a hundred percent sure, but I tried to keep the behavior we had with the previous version that was returning all collections. It took me some time to get the logic of this part so maybe I didn't get it right, so here is my thinking:

The change happened in the "all_collections" function of stac-fastapi-pgstac

In the previous version, (here https://github.com/stac-utils/stac-fastapi-pgstac/blob/b343902d243bd50048daaa27263488b8f72f82eb/stac_fastapi/pgstac/core.py#L113), the query we mapped was "SELECT * FROM all_collections();", which was returning all collections, but in the code we can notice that the query "SELECT * FROM collection_search($1::text::jsonb);" was already in use but in a specific "if" (if the extension CollectionSearchExtension is enabled). As we didn't have this query mapped in our code with this version, I believe this is a case we don't support.

In the new version (here https://github.com/stac-utils/stac-fastapi-pgstac/blob/4b0ea125cfd4650447eeaac35bd77f4fd0f175e7/stac_fastapi/pgstac/core.py#L102), there is no more the "all_collection" query, everything goes to the "collection_search" query, but there is still an if/else with a condition on CollectionSearchExtension that decides if we use advanced parameters (in case it's enabled) or basic ones (in case it's not). As in the previous version we didn't handle the case when it was enabled, I kept this logic that we only care about the basic case. And in the basic case, the parameters don't contain args that allow filtering collections, so it always return all collections.

As I said I'm not 100% sure about this thought process, I've never worked with part before, so let me know if I'm wrong or if I can improve things.

vprivat-ads: Hi @ecombelles-csgroup yes we should enable the [collections-search](https://github.com/stac-api-extensions/collection-search) extension now that stac-fastapi supports it. It will allow us to use this API from stac browser. We must still keep our filtering based on user access control. If I search for sentinel-1 collections I shall retrieve the sentinel-1 collections for which I have access, no more no less.

vprivat-ads (approved):

jgaucher-cs (approved):

